Issue #4, opened 2014-10-09
Provide a choice of esprima versions or even parsers


fkling:
Possible other parsers:
- https://github.com/marijnh/acorn

Also provide a way to change parsing options if available.

fkling:
This was kind of done in 5f7fa2d049df55b03d54965e02207292f0196e65. I might add acorn, but I'll close this for now.

PR #13, opened 2015-05-06
Include comments in AST

kaustubh-karkare:


fkling:
Cool! Can you please include the minified version of `app.js`? I.e. the one generated by `npm run build`.

hzoo:
@fkling Are you planning on making babel the default on github pages (http://felix-kling.de/esprima_ast_explorer/)? I do see #4

fkling:
@hzoo: Good question. I'm planning to create a new fork of the tool which would allow selecting different parsers / languages, over here: https://github.com/fkling/ast-explorer. Alas, I haven't found the time to get around it yet.

However, as a first step, I could simply move the babel branch over there. I don't really want to make Babel the default here, since after all this is named "esprima_ast_explorer" ;) (and renaming it might be confusing).

PR #27, opened 2015-08-24
add babel-eslint parser, using acorn-to-esprima pacakge for smaller bundle

### LINKED ISSUE #20: Add more JS parsers
- [x] babel-eslint
- [x] espree
- [x] acorn
- [x] recast
- [x] shift-ast

hzoo:
Fixes #20 

Am I supposed to commit the `dist/` files as well? I removed them for now.

screen shot 2015-08-24 at 2 27 10 pm

I don't know if we want to filter out the `_paths` property or not

### DIFF
--- a/src/parsers/babel-eslint.js
+++ b/src/parsers/babel-eslint.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import pkg from 'acorn-to-esprima/package.json';
+import loadAndExecute from './utils/loadAndExecute';
+
+const ID = 'acorn-to-esprima';
+const name = 'babel-eslint';
+
+export default {
+  id: ID,
+  displayName: name,
+  version: pkg.version,
+  homepage: pkg.homepage,
+
+  parse(code) {
+    return loadAndExecute(
+      ['acorn-to-esprima', 'babel-core'],
+      (acornToEsprima, {acorn, traverse, parse}) => {
+        const opts = {
+          locations: true,
+          ranges: true
+        };
+
+        const comments = opts.onComment = [];
+        const tokens = opts.onToken = [];
+
+        let ast;
+        try {
+          ast = parse(code, opts);
+        } catch (err) {
+          throw err;
+        }
+
+        ast.tokens = acornToEsprima.toTokens(tokens, acorn.tokTypes);
+        acornToEsprima.convertComments(comments);
+        ast.comments = comments;
+        acornToEsprima.attachComments(ast, comments, ast.tokens);
+        acornToEsprima.toAST(ast, traverse);
+
+        // remvove out node._paths
+        traverse(ast, {
+          noScope: true,
+          exit: function (node) {
+            if (node._paths) {
+              delete node._paths;
+            }
+          }
+        });
+        delete ast._paths;
+
+        return ast;
+      }
+    );
+  },
+
+  nodeToRange(node) {
+    if (typeof node.start !== 'undefined') {
+      return [node.start, node.end];
+    }
+  }
+};
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ The JavaScript AST explorer provides
 - [acorn][]
 - [babylon][]
 - [recast][]
+- [shift][]
+- [babel-eslint][]
 
 to parse the code. Depending on the parser settings, it not only supports ES5
 but also
@@ -21,7 +23,7 @@ but also
 
 Since future syntax is supported, the JavaScript AST explorer is a useful tool
 for developers who want to create AST transforms.
-In fact, [jscodeshift][] and [babel][] are included so you can prototype 
+In fact, [jscodeshift][] and [babel][] are included so you can prototype
 codemodding scripts and babel plugins.
 
 ### Features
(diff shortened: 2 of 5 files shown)


hzoo:
@fkling is "Add the module to parition-bundle.js (either as new bundle or to an existing one of there are cross-dependencies)." now `partition.json`? and I guess I should update the shrinkwrap too

fkling:
> is .... now `partition.json`?

Oh yeah, I wrote that without verifying the file name ;)

> I don't know if we want to filter out the _paths property or not

I assume people don't usually access it? In that case yes. Still have to think about a good way for parsers to specify fields to hide.

> Am I supposed to commit the `dist/` files as well?

Nope, I will do the rebuild.

> and I guess I should update the shrinkwrap too

I can do that too.

---

`babel-lint` itself loads a bunch of dependencies that are not necessary in the browser. Can we just use https://github.com/babel/babel-eslint/blob/master/acorn-to-esprima.js directly with the existing babel-core parser? Or would that not work as desired?

hzoo:
> Can we just use https://github.com/babel/babel-eslint/blob/master/acorn-to-esprima.js directly with the existing babel-core parser? Or would that not work as desired?

I wanted to try doing that - most of `index.js` is patching escope which isn't needed for the ast. I was going to split acorn-to-esprima to another repo but never got around to it yet https://github.com/babel/acorn-to-esprima

Yeah I think we'd only need https://github.com/babel/babel-eslint/blob/master/index.js#L445-L462 + acorn-to-esprima

src/parsers/babel-eslint.js
```diff
@@ -0,0 +1,28 @@
...
+    );
+  },
+
+  nodeToRange(node) {
+    if (typeof node.start !== 'undefined') {
+      return [node.start, node.end];
+    }
+  },
+
+  renderSettings() {
```
fkling: Just don't provide the method and the settings button will be disabled.

hzoo: K wasn't sure :+1:

hzoo:
Yeah we'll need to filter `_path` since it's on every node and results in `Uncaught RangeError: Maximum call stack size exceeded`

fkling:
Do you know where exactly? Might be in `getFocusPath`.

hzoo:
https://github.com/eslint/eslint/issues/3440 was doing a `for (var prop in node) {` check so something like that

hzoo:
OK haha easy find - yeah `getFocusPath` https://github.com/fkling/esprima_ast_explorer/blob/e638e420140515b51141c2d6bbe1ad58879e6c57/src/getFocusPath.js#L32

fkling:
> I think we should just remove that property on the babel-eslint side?

I don't know, do you need it? :) Otherwise I would have to find a way for the parser to let the system know which properties to ignore.

> Yeah I think we'd only need https://github.com/babel/babel-eslint/blob/master/index.js#L445-L462 + acorn-to-esprima

Can we [factor this out](https://github.com/babel/babel-eslint/blob/master/index.js#L416-L462) into its own function / method and only require that one?

hzoo:
I've never had to use it so probably not :stuck_out_tongue: - I'l try to integrate the relevant code into an acorn-to-esprima package. I guess I should move the tests as well and all that.

hzoo:
Ok made https://github.com/babel/acorn-to-esprima. Getting an odd

screen shot 2015-08-25 at 12 08 47 pm

I'l push my changes soon

hzoo:
Yeah maybe you can take a look

fkling:
Mmh, I've seen this sometimes when the editor does receive a string as value. Will have a closure look later.

hzoo:
Ok I was dumb - after I made acorn-to-esprima I still though I was using babel-eslint still - so I had to copy paste the parse method from there? Otherwise I would of just used it like I tried intially.

Still need to remove the `_paths` but otherwise works as well

hzoo:
Ok updated - no babylon (since babel-core has the parse method) and loaded babel-core dynamically. Thanks guess it was unsure of how it was working.

hzoo:
Ok able to remove `node._paths`

I guess the options could be created in the loadAndExecute callback - if options is true, then remove tokens, comments, etc.

hzoo:
K both PR should be working now - I added npm-shrinkwrap for this one so see if it's working locally for you.

src/parsers/babel-eslint.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import pkg from 'acorn-to-esprima/package.json';
+import loadAndExecute from './utils/loadAndExecute';
+
+const ID = 'acorn-to-esprima';
+const name = 'babel-eslint';
```
hzoo: Since people don't know what acorn-to-esprima is - the dropdown option will be called `babel-eslint` but the version is `acorn-to-esprima-1.0.2` rather than the wrong `babelEslint-1.0.2` if that makes sense.

fkling:
Merged with 0f3a1ebceeb8462db1f3b0dd02c9a4bf58cd401f !

Issue #26, opened 2015-08-24
Add more tools


fkling:
There are many more tools besides jscodeshift which could be added to the ast explorer for quick prototyping / testing. For example:

- eslint plugin development
- babel plugin development

To make adding new tools easy, some refactoring of the client side code and the DB schema is necessary. I already have a proof of concept commit locally but it still needs more tweaking.

The current idea/plan is that "tools" are React components which are rendered below the code editor and ast panel (just like the transform tool now) and which have access to

- The current parser
- The current code
- An API to highlight parts and show errors in the code editor

Tools would also be able to specify which parsers they work together with. They could render whatever interface they need.

hzoo:
:+1: been using this for jscs and babel-eslint. Can be a lot of work to debug/make rules/tests.

jasonLaster:
:+1:

fkling:
First step is done: The logic is more split up to allow custom transformers (but not custom views yet).

jasonLaster:
nice! can't wait to see that

fkling:
@jasonLaster: If you go to http://astexplorer.net/, you can now select between "jscodeshift" and "babel".

fkling:
We have support for PostCSS now.

fkling:
I'm closing this for now, since we have everything that I originally wanted. In the future, we might also want to give more liberty to tools and let them render whatever they want in their panel (i.e. they wouldn't be restricted to just two text input fields).

PR #29, opened 2015-09-01
Plugin architecture for the transform panel, add babel live coding support

cpojer:
See #26 

@gaearon requested it here: https://twitter.com/dan_abramov/status/638659321228840960

Should have built it like this from the beginning.

There are some minor annoyances with toggling transformers – how do we alert the user that toggling means they have to rewrite their code? I made it so it only wipes away the default transform code but the code there is slightly messy. I'm sure @fkling can make it much much better.

Also probably needs an update so that Parse saves the used transform plugin.


gaearon:
This. Is. So. Rad! I love it.

I see a dropdown glitch in Chrome:

screen shot 2015-09-01 at 21 46 32

cpojer:
Hey now, don't hate on my CSS.

src/transformers/jscodeshift.js
```diff
@@ -0,0 +1,37 @@
...
+const defaultTransform =
+  fs.readFileSync(__dirname + '/transformJscodeshift.txt', 'utf8');
+
+function transform({transformCode, code}) {
+  return new Promise((resolve, reject) => {
+    loadjs(['babel-core', 'jscodeshift'], (babel, jscodeshift) => {
+      try {
+        // This might throw
+        var transform = compileModule(
+          babel.transform(transformCode).code
```
cpojer: I felt really tempted to self-host the plugin with a previous version of the transformCode, like

```
babel.transform(transformCode, {plugins: [previousTransformVersion]});
```

:lollipop:

@sebmck

fkling:
I guess this is a good first strep. My actual idea was to abstract even more so that "tools" can specify their own view. E.g. for eslint stuff it will be necessary to have a panel to define options.

However simple transform tools could build on this. And yeah, I have to update the parse code as well.

fkling:
Can you just not commit the `dist/` folder? I will rebuild anyway.

src/transformers/utils/compileModule.js
```diff
@@ -0,0 +1,6 @@
+export default function compileModule(code) {
+  var m = {};
+  var f = new Function('module', `(function() {var exports = {}; ${code}}())`);
```
DmitrySoshnikov: Can you assign to the local `exports`? How is it returned?

cpojer: `module.exports = exports;` is added by babel.

cpojer:
@fkling got rid of the dist folder from the commit. Am I not supposed to commit it like I used to?

fkling:
I will rebuild anyway.

PR #33, opened 2015-09-04
Allow running offline

forivall:
I was using my fork a month or so ago without internet access. Could use indentation fixes, but not indenting makes it easier for me to merge.


fkling:
I think this isn't needed anymore, given that we use the Parse npm package now.

PR #37, opened 2015-10-07
Allow user selection of strings in tree

zertosh:
Mostly so you can copy+paste node types. cc: @fkling

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -305,6 +305,15 @@ li.entry {
   background-color: rgba(255,240,6,0.4);
 }
 
+.entry > .value .s {
+  cursor: text;
+  -webkit-user-select: text;
+  -khtml-user-select: text;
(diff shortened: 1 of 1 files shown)


pieterv:
So happy about this! Nice work @zertosh

fkling:
Nice! I disabled selection, because quickly clicking on the togglers would select the text otherwise, which is annoying. But enabling this for values makes absolutely sense.

How about adding

```css
cursor: text;
```

as well, so that its clear that this is selectable?

zertosh:
Good idea! I rebased the change

hzoo:
:+1:

fkling:
Thanks so much!

fkling:
Updated `gh-pages` branch (373dfbabc5dff033e1cbdd1850f683acb247a9c4).

PR #39, opened 2015-10-20
Guard against infinite loops

ForbesLindesay:
This guards against infinite loops using the halting-problem module.

It catches simple cases like the following statically:

```js
export default function ({Plugin, types: t}) {
  return new Plugin('ast-transform', {
    visitor: {
      Identifier(node) {
        var n = 100;
        while (true) n++;
        return t.identifier(node.name.split('').reverse().join(''));
      }
    }
  });
}
```

results in

> while statement on line 6 has a constant test.  Either it is always `false` (making the statement redundant) or it is an infinite loop.

It can catch more subtle infinite loops like the following by checking for a timeout (of say 5 seconds) in all loops.  In this case it will just give you a message of "Infinite loop detected on line 6"

```js
export default function ({Plugin, types: t}) {
  return new Plugin('ast-transform', {
    visitor: {
      Identifier(node) {
        var n = 100;
        while (n > 0) n++;
        return t.identifier(node.name.split('').reverse().join(''));
      }
    }
  });
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "escodegen": "^1.4.1",
     "espree": "^2.2.4",
     "esprima": "^2.5",
+    "halting-problem": "^1.0.1",
     "jscodeshift": "^0.3",
     "keypress": "git://github.com/dmauro/Keypress",
     "pubsub-js": "^1.4.2",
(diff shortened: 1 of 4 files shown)


fkling:
This is great, thank you!

kyldvs:
Awesome!

cpojer:
This is awesome, thank you @ForbesLindesay !

PR #46, opened 2015-11-09
Add TypeScript support

nmalaguti:


JKillian:
:+1:

fkling:
Awesome, thank you, I will give it a try later.

src/parsers/typescript.js
```diff
@@ -0,0 +1,126 @@
...
+            }
+          }
+        };
+
+        const transformNode = (node) => {
+          if (typeof node === 'object' && node.kind) {
+            if (typeof node.type !== 'undefined') {
+              node._typescriptType = node.type;
+            }
+            node.type = ts.SyntaxKind[node.kind];
```
fkling: Instead of treating "parent" special everywhere else, could we simply remove it here?

nmalaguti: Sadly, if you remove `parent` the node.getStart() and node.getEnd() methods stop working properly in the nodeToRange function below :(.

fkling:
FWIW, looking at this really makes me want to move traversing and "serialization" logic into the parser. I.e. the parser should have methods that, given an AST node, returns the short name, a renderable representation and a list of children. That would make keep parser / AST specific logic together and may allow us to avoid traversing the whole AST to prepare it for output.

I might work on that in the next couple of days.

src/parsers/typescript.js
fkling: Oh I see, OK, no worries. See my other comment, that should allow us to get rid of AST specific logic all over the place.

fkling:
Merged with 1e26e5dd63770b40680f41d84d5c34eb62999cae (I squashed the commits)

PR #48, opened 2015-11-12
Fix method Set uses for membership

nmalaguti:
Set uses `has` not `contains` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has).

It looks to be related to change: https://github.com/fkling/esprima_ast_explorer/commit/4ebc1fb119452479d8623c0085cefb9847e2ac0f#diff-88541fdf6f0855d0c9f2f9c3a51c48c2R35.

### DIFF
--- a/src/getFocusPath.js
+++ b/src/getFocusPath.js
@@ -32,7 +32,7 @@ export default function getFocusPath(node, pos, parser, path) {
     if (
       prop !== 'range' &&
       prop !== 'loc' &&
-      !ignoreProperties.contains(prop) &&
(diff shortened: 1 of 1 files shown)


hzoo:
Ah yeah and that change was due to https://github.com/babel/acorn-to-esprima/pull/2

fkling:
Oops. Wrote too much Hack lately. Thanks for noticing!

@hzoo, I see. Then we wouldn't even need to include `__path` here, but I'm going to change all this anyway (give more control to the parser for how to interpret AST nodes).

PR #84, opened 2015-12-13
Add a GraphQL parser

fson:
- Add GraphQL category with graphql-js parser.
- Set a default icon for categories without a Font Awesome icon.

screen shot 2015-12-13 at 16 41 25


fkling:
Very nice! Looks good to me, will merge later. We may not keep the "GraphQL" category, but it's fine for now.

PR #87, opened 2015-12-22
Add ability to hide location data

yairhaimo:
Added the ability to hide the node's location data in order to remove clutter when interrogating the structure and data of the node.
Location property names are configurable per parser.

### DIFF
--- a/src/parsers/css/cssom.js
+++ b/src/parsers/css/cssom.js
@@ -10,7 +10,7 @@ export default {
   displayName: ID,
   version: pkg.version,
   homepage: pkg.homepage,
-  locationProps: ['__starts', '__ends'],
+  locationProps: new Set(['__starts', '__ends']),
 
   loadParser(callback) {
     require(['cssom/lib/parse'], callback);
--- a/src/parsers/css/cssom.js
+++ b/src/parsers/css/cssom.js
@@ -10,6 +10,7 @@ export default {
   displayName: ID,
   version: pkg.version,
   homepage: pkg.homepage,
+  locationProps: ['__starts', '__ends'],
 
   loadParser(callback) {
     require(['cssom/lib/parse'], callback);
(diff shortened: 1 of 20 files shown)


RReverser:
Thanks. Idea is probably reasonable. Looks like you didn't update `css`/`html` parsers though?

yairhaimo:
Hmm, true. Right on it.

RReverser:
Also - maybe we should hide `type`-related fields as well? @fkling what do you think?

yairhaimo:
In regards to postcss - I opted to remove only the 'start' and 'end' inside 'source' and not 'source' itself because I dont know if it can contain more properties.

RReverser:
Nah, `source` is just already-hidden `input` source + locations within it, so it's safe to hide completely under this option.

yairhaimo:
Ok, sorry for missing yet another parser. Added graphql and changed postcss locationProps to 'source'.

fkling:
Great, thank you! That's also something I wanted for a while.

@RReverser:
> maybe we should hide `type`-related fields as well

Maybe we can consolidate this option into something like "Hide meta data", which includes location and type information. But then we would probably discuss all day what should be meta data and what shouldn't.

src/parsers/css/cssom.js
```diff
@@ -10,6 +10,7 @@ export default {
   displayName: ID,
   version: pkg.version,
   homepage: pkg.homepage,
+  locationProps: ['__starts', '__ends'],
```
fkling: How about using a `Set` here instead? I know it will probably introduce more overhead but it would be more in line with `_ignoredProperties`.

RReverser:
> Maybe we can consolidate this option into something like "Hide meta data", which includes location and type information. But then we would probably discuss all day what should be meta data and what shouldn't.

Personally, I'm fine with that name for type and locations (anything that is already represented in UI in another form).

yairhaimo:
I just changed the array to a Set but didnt touch the Type property or any other meta data.
Let me know if you want me to add it and change the feature to "hide meta data".

fkling:
We can always iterate on it.

Issue #89, opened 2015-12-27
can't run locally


kolodny:
I ran `npm install` and either `npm run watch` or `npm run build` and in another tab terminal I run `npm start` yet this is the contents of index.html:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>AST explorer</title>
    <meta charset="UTF-8">
    <meta name="description" content="An online AST explorer.">
    <script>
      window.webpackManifest = {%# o.manifest %}
    </script>
  </head>
  <body>
    <div id="page">
      <div id="container"></div>
      <div id="contribution">
        Built using
        <a href="http://facebook.github.io/react/">React</a>,
        <a href="http://babeljs.io/">Babel</a>,
        <a href="http://fortawesome.github.io/Font-Awesome/">Font Awesome</a>,
        <a href="https://parse.com">Parse.com</a>,
        <a href="http://codemirror.net/">CodeMirror</a>
        and
        <a href="https://webpack.github.io/">WebPack</a>
        |
        <a href="https://github.com/fkling/esprima_ast_explorer">GitHub</a>
      </div>
    </div>
  </body>
</html>

```

it looks like it's not modifying `window.webpackManifest = {%# o.manifest %}`

I'm on mac os with node v5.1.0 and npm 3.3.12
I'm also getting the same issue on node v4 too

PR #90, opened 2015-12-27
serve correct directory

kolodny:
see https://github.com/fkling/astexplorer/issues/89


fkling:
Thanks! I missed that...

PR #93, opened 2016-01-21
Add flow-parser

jeffmo:
`flow-parser` is a compiled-to-JS version of the OCaml Flow parser. It's mildly out of date right now, but I'll follow up with a version bump once we ship Flow 0.21 in the next few weeks.

Tested this by building locally, loading up, trying out the new parser. Settings tab is grayed out for now -- parsing works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The AST explorer provides following code parsers:
   - [traceur][]
   - [typescript][]
   - [uglify-js][]
+  - [flow-parser][]
 - CSS:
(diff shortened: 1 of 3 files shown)


hzoo:
I would add an entry to the README too! (other than the package.json change)

jeffmo:
doh! updated package.json (and the readme)

fkling:
Sweet!

fkling:
Updated the site.

PR #96, opened 2016-02-06
Fix bug changing Babylon sourceType

btmills:
The `changeOption` handler was attempting to read from `target.vaue` instead of `target.value`. Previously, changing `sourceType` from "module" (the default) to "script" would cause the parser to warn about module-specific code, but reopening settings would show that `sourceType` was still "module".


hzoo:
Nice catch haha

fkling:
Thanks a lot!

PR #98, opened 2016-02-15
Fix Espree ecmaVersion setting

btmills:
Previously, `ecmaVersion` was being set as a string, but Espree expects a number. This change ensures that `ecmaVersion` is always set as a number.


fkling:
Whoops, thanks!

PR #104, opened 2016-03-21
Add new transformer for creation of ESLint rules

Daniel15:
One of the most common ways I use ASTExplorer is while creating new ESLint rules. This pull request allows creation of ESLint rules in ASTExplorer. The ESLint rule is entered as a "transform", and its results are displayed in the output pane:

astexplorer

Live demo: http://stuff.dan.cx/astexplorer/


Daniel15:
One of the issues I've seen is that the ESLint Webpack bundle is giant. Might need to figure out what stuff isn't needed and why it's being pulled in, or add some more `webpack.IgnorePlugin`s to the Webpack config. It's working apart from that though!

hzoo:
:tada: Oh nice! This is so awesome @Daniel15 (Ref https://github.com/fkling/astexplorer/issues/26)

src/parsers/js/transformers/eslint/index.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+function formatResults(results) {
+  return results.length === 0
+    ? 'Lint rule not fired.'
+    : results.map(formatResult).join('').trim();
+}
+
+function formatResult(result) {
+  var pointer = ' '.repeat(result.column - 1) + '^';
```
hzoo: Thoughts on using `-` instead of a space to do `----^`? We do this in jscs

Daniel15: That sounds reasonable :) I'll update it.

Daniel15:
I'm still optimising the Webpack config a bit. There's a bunch of random junk being pulled into the Webpack bundle due to some dynamic requires in ESLint, so it's bundling the entire ESLint directory rather than just the stuff that's needed. I'm manually going through it and figuring out which modules can be ignored or shimmed in the Webpack config, and how to stop it doing dynamic requires (https://github.com/webpack/webpack/issues/198)

RReverser:
It's amazing, thanks for working on this!

fkling:
Could you squash the commits please? Otherwise I will just do it when I merge :)

hzoo:
Yeah I don't know what we can do to update to eslint 2 yet due to babel-eslint issues

Daniel15:
> Could you squash the commits please? 

Sure, I'll do that tonight.

>Yeah I don't know what we can do to update to eslint 2 

Interesting, I didn't even realise ESLint 2 existed. All I did was `npm install --save eslint` which installed ESLint 1 (maybe due to the version of `babel-eslint` that's installed?)

RReverser:
@hzoo What kind of issues that we might face in astexplorer?

hzoo:
Haha I just assumed since you didn't run the patching or anything (that would just mean that if you make a rule with es2015+ or flow it might not work correctly).

It might work, but we would need to update the babel-eslint version to 6.0.0-beta.1 to support it.

Daniel15:
@fkling - I spent a while messing around with Git and still can't figure out how to square the commits. I've never squashed commits before. Sorry :/ Maybe you'll have better luck than me if you squash them when you merge.

hzoo:
For future reference, doing an interactive rebase would work. `git rebase -i HEAD~7` which would bring up a window with the past 7 commits.

```
pick 1302f5e Upgrade Acorn and Espree for ES7 support
pick 872e959 Add new transformer for creation of ESLint rules
pick 234dc98 Use hyphens rather than spaces for arrow pointing to line of code, and ignore a bunch of unnecessary modules to reduce bundle size.
pick cb966d7 Optimize Webpack config by disabling dynamic requires for ESLint files (unnecessary when using it the way we're using it). It's now much smaller.
pick 34780b6 Remove unnecessary "options" object (copypasta fail)
pick 827514a Include changes I forgot to commit
pick eb0db99 Rename to eslint1

# Rebase 5d74436..eb0db99 onto 5d74436 (7 command(s))
```

Then you can replace the bottom 6 commits' `pick` word into fixup/squash and it will do it.

Or `git reset --soft HEAD~6` which resets the last 6 commits and lets you run `git commit` again

src/parsers/js/transformers/eslint/index.js
```diff
@@ -0,0 +1,66 @@
...
+function formatResults(results) {
+  return results.length === 0
+    ? 'Lint rule not fired.'
+    : results.map(formatResult).join('').trim();
+}
+
+function formatResult(result) {
+  var pointer = '-'.repeat(result.column - 1) + '^';
+  return `
+\u2022 ${result.message} (at ${result.line}:${result.column})
```
RReverser: Maybe better to use `//` here so that JS highlighter wouldn't treat that line as a code but rather as a comment?

hzoo: Same with the pointer? I think it be ok as it is, but it would look like

screen shot 2016-03-22 at 9 45 13 am

RReverser: Yeah, agreed. Would be nice to have proper highlighting and be sure that "keywords" won't be attempted to be highlighted.

hzoo: screen shot 2016-03-22 at 9 59 54 am

I see (so wrapping `'` could work too), and maybe it's not as necessary for the pointer

package.json
```diff
@@ -49,6 +49,7 @@
     "cssom": "^0.3.0",
     "domhandler": "^2.3.0",
     "escodegen": "^1.4.1",
+    "eslint": "^1.10.3",
```
hzoo: Hm since we use eslint in the devDependencies, do we want to keep it there or do we want to remove that?

fkling: I think, in the long term, we need to find a better solution to include parsers / tools which also happen to be used for building astexplorer itself.

src/parsers/js/transformers/eslint/index.js
RReverser: With wrapping into `'` you need to care about internal possible `'` in error messages. I would rather go with `//` as it can be terminated only by line terminator :)

Daniel15: Is there a way to disable syntax highlighting in the output pane? I'd
rather just totally disable syntax highlighting for ESLint results as it
doesn't really make sense.

RReverser: Why so? It makes perfect sense for code slices IMO.

hzoo:
Yeah it would be cool to create an interface to allow adding test cases or something and generate a complete eslint rule file + tests - as something you can put into eslint or a plugin easily.

package.json
hzoo: Is it possible to use something like http://www.webpackbin.com/ or tonicdev to be able to choose versions (might not be directly related to this issue)?

src/parsers/js/transformers/eslint/index.js
Daniel15: @RReverser  - Because the code could be a partial snippet that might break the syntax highlighting. Consider the trivial "Do not use template literals" rule I included as an example. If you have a multi-line template literal, the syntax highlighting would break due to the fact that the template literal is not closed:

screen shot 2016-03-22 at 8 11 25 pm

(see the orange carat)

I don't think it's safe to syntax highlight partial code snippets unless they're all individually highlighted.

fkling:
> Yeah it would be cool to create an interface to allow adding test cases or something and generate a complete eslint rule file + tests - as something you can put into eslint or a plugin easily.

My plan always was that "transforms" (I'd rather call them "tools") can completely define their own interface, instead of just having a fixed input and output panel. That would allow UIs that are more specialized towards each individual tool. I'm happy to review any contribution in this direction :)

package.json
RReverser: Or we might just use separate repo with purely parser API and deps, which will be referenced as a dependency by UI.

hzoo:
Sounds good. What if you could run a babel/eslint plugin on your own code and link it from on astexplorer - like in https://twitter.com/rauchg/status/712800259089391617 :smile: - either that or it creates a repo/files.

fkling:
> link it from on astexplorer 

That will be easier to do once we switch to a custom server and can provide an API for reading code transforms. Or as @RReverser suggested in the context of jscodeshift, we could switch to storing code snippets in gists instead of Parse.

Daniel15:
Okay I think I squashed it properly. I was trying to do it through my Git GUI (SourceTree, on Windows) and couldn't figure it out so I used `git rebase -i` like @hzoo suggested :)

hzoo:
Cool - looks good, I see we are going with the comments for the output.

I just saw a tweet about this too https://twitter.com/domenic/status/713417333549764608.

Daniel15:
I wish Github was as good as Phabricator... Phabricator automatically squashes the diff when you merge the change. Maybe one day Github will actually have decent code review tools.

RReverser:
@Daniel15 It's even easier to do in SourceTree - right-click last commit before any of yours, in the dialog click "Combine with previous" couple of times until all of your commits are grouped. Then Edit message, set to reflect thr summary of changes and done.

Daniel15:
@RReverser - I tried that, and it didn't work properly. It only seemed to include some of the commits, not all of them. Anyways, I know how to do it through the command line now, which I suppose is useful to know :D

fkling:
Just a quick clarification question: This currently uses eslint v1 to run the plugin. Is plugin development still the same in eslint v2?


@hzoo 
> Yeah I don't know what we can do to update to eslint 2 yet due to babel-eslint issues

I guess we should be able to use eslint's internal parser now? Ideally we would have a configuration dialog for tools as well, where one could select the parser.

hzoo:
It's mostly the same http://eslint.org/docs/user-guide/migrating-to-2.0.0 - there are a few changes like different rule schemas, a config change (not plugin related necessary), but the main stuff is the same

fkling:
I'd say we should try to get this to work with eslint v2. Without having to mess with astexplorers own eslint stuff, we can add eslint v2 similar to how I added babel v6: https://github.com/fkling/astexplorer/tree/master/packages .

Daniel15:
We're still using ESLint 1 at Facebook internally which is why I focused mainly on that (that, and it was the version ASTExplorer was already pulling in. I can update it to ESLint v2 tonight if I get some free time. Thanks for the link to the packages folder, I didn't notice that at first.

fkling:
> We're still using ESLint 1 at Facebook internally

Oh, I didn't know that. In that case, we could also do same we do with Babel and offer `eslint v1` and `eslint v2`. So this PR would the the v1 version and we can add a v2 version later.

hzoo:
Sounds good to me (v1 and v2)

Daniel15:
Cool, in that case I think we should get this version in in its current state (with ESLint v1) and then add ESLint v2 in a followup pull request :smile:

fkling:
Yes! I will probably rename the ID to `eslintv1`.

hzoo:
You can ping me if you need help with the next PR @Daniel15

fkling:
https://twitter.com/fkling42/status/715254117745070080

Daniel15:
Thanks!

It looks like I was wrong with this:
> We're still using ESLint 1 at Facebook internally

Seems like we upgraded to ESLint v2 very recently. It was using ESLint v1 when I first sent this pull request, but I checked again today and it's v2 now :stuck_out_tongue: I'll try to send a PR adding ESLint v2 support tonight.

fkling:
lol

>  I'll try to send a PR adding ESLint v2 support tonight.

Awesome, thank you!

PR #105, opened 2016-03-25
Add acorn-loose parser

TimothyGu:


src/parsers/js/acorn-loose.js
```diff
@@ -0,0 +1,94 @@
...
+  displayName: ID,
+  version: `${pkg.version}`,
+  homepage: pkg.homepage,
+  locationProps: new Set(['range', 'loc', 'start', 'end']),
+
+  loadParser(callback) {
+    let _localReq = require;
+    require(['acorn'], (acorn) => {
+      global.acorn = acorn;
+      _localReq(['acorn/dist/acorn_loose'], callback);
```
TimothyGu: The `acorn/dist/acorn_loose.js` file is built with Browserify, which means that the Webpack `require` function cannot get in there. Hence these hacks.

fkling:
I feel like this should rather be an option in the acorn parser settings, and based on that option it would choose which parser to use. Other than avoiding to load the loose parser if it's not needed, I don't see an advantage to have it as a separate entry.

Thoughts?

hzoo:
Yeah, I think that would be good even though its a separate file/etc. https://github.com/ternjs/acorn#distacorn_loosejs

TimothyGu:
@fkling @hzoo Done.

src/parsers/js/acorn.js
```diff
@@ -24,16 +24,33 @@ export default {
   locationProps: new Set(['range', 'loc', 'start', 'end']),
 
   loadParser(callback) {
+    const parsers = {};
+    const _localReq = require;
+
     require(['acorn', 'acorn-jsx/inject'], (acorn, jsxInject) => {
```
fkling: I think we should be able to directly require `src/loose/index.js` here and avoid all the issues with loading the built version?

fkling: Also, we should probably do the same thing with acorn itself, but that doesn't have to be done in this PR.

TimothyGu: Fixed.

PR #107, opened 2016-03-31
Add support for ESLint v2

Daniel15:
I copied how Babel 6 is handled - Created "modules" under `packages` which export the correct version of ESLint. Moved shared code into a `eslintUtils` file. 

Pretty much all the transform logic is identical across ESLint 1 and 2. I kept the transformers themselves (`eslint1/index.js` and `eslint2/index.js`) separate rather than totally sharing them (eg. a `createESlintTransformer` factory method) so that it's easier for them to deviate if needed.


Daniel15:
Hopefully the npm package stuff is okay, I kept getting strange permission errors like this when running `npm install`:

```
npm ERR! Windows_NT 10.0.10586
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
npm ERR! node v5.7.1
npm ERR! npm  v3.6.0
npm ERR! path C:\src\astexplorer\node_modules\eslint2
npm ERR! code EPERM
npm ERR! errno -4048
npm ERR! syscall rename

npm ERR! Error: EPERM: operation not permitted, rename 'C:\src\astexplorer\node_modules\eslint2' -> 'C:\src\astexplorer\node_modules\.eslint2.DELETE'
npm ERR!     at moveAway (C:\Program Files\nodejs\node_modules\npm\lib\install\action\finalize.js:38:5)
npm ERR!     at destStatted (C:\Program Files\nodejs\node_modules\npm\lib\install\action\finalize.js:27:7)
npm ERR!     at FSReqWrap.oncomplete (fs.js:82:15)
npm ERR!
npm ERR! Error: EPERM: operation not permitted, rename 'C:\src\astexplorer\node_modules\eslint2' -> 'C:\src\astexplorer\node_modules\.eslint2.DELETE'
npm ERR!     at Error (native)
npm ERR!  { [Error: EPERM: operation not permitted, rename 'C:\src\astexplorer\node_modules\eslint2' -> 'C:\src\astexplorer\node_modules\.eslint2.DELETE'] parent: 'astexplorer' }
npm ERR!
npm ERR! Please try running this command again as root/Administrator.
```

It resolved itself if I ran the command again. I think it's just npm being glitchy and not an actual issue. (I've seen the same thing on Linux and on Windows so I don't think it's an OS-specific issue)

src/parsers/js/transformers/eslint1/index.js
```diff
@@ -0,0 +1,23 @@
+import {defineRule, runRule} from '../../utils/eslintUtils';
```
fkling: The problem with this is that the transitive dependencies, `babel-eslint` and `babel-core` are now loaded on every page load, not only when someone uses the eslint transform.

You could simply load the utils dynamically as well.

src/parsers/js/transformers/eslint2/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+  defaultParserID: 'acorn-to-esprima',
+
+  loadTransformer(callback) {
+    require(['eslint2'], callback);
+  },
+
+  transform({ eslint, rules, sourceCode }, transformCode, code) {
+    defineRule(rules, transformCode);
+    return runRule(code, eslint, sourceCode);
```
fkling: @hzoo: What exactly are the issues with eslint2 and babel-eslint? Should we just not use babel-eslint here?

fkling:
Thank you so much for doing this!

src/parsers/js/transformers/eslint1/index.js
Daniel15: Oh, good catch! I thought Webpack would automatically make a new bundle for every `transformers/*/index.js` but maybe it doesn't work that way. How does the code splitting work? It is simply the `require([..], cb)` that does it? I've never used that feature of Webpack.

fkling: Yep, the `require([..], cb)` does the magic :)

src/parsers/js/transformers/eslint2/index.js
Daniel15: I had trouble getting espree to run properly (it kept giving me syntax errors for some reason) so I just gave in and used babel-eslint like the ESLint 1 transformer does. I guess it doesn't really matter which parser is used as ESLint can just use the AST.

hzoo: Well the issue has to do with how babel-eslint monkey patches ESLint - however we aren't running that so there aren't any issues (parseNoPatch instead of parse). This just means certain things won't work correctly like a scope rule dealing with flow types etc.

Daniel15: I need to use `parseNoPatch` since the monkey-patching doesn't work within the context of a browser. `Module._resolveFilename

Daniel15: @hzoo - I needed to use `parseNoPatch` since the monkey-patching doesn't work within the context of a browser (or really any non-Nodejs environment), as `Module._resolveFilename` `new Module` and ` Module._nodeModulePaths` are Node-isms. It assumes a Node.js environment.

Daniel15:
Updated the diff to also dynamically require `eslintUtils`.

PR #109, opened 2016-04-03
Add proper Web IDL highlighting

TimothyGu:
Before the CodeMirror IDL mode was used, which was not for Interface Definition Language but rather for the unrelated Interactive Data Language.

The Web IDL mode for CodeMirror, included in this PR, is adapted from the mode submitted to CM as codemirror/CodeMirror#3933.

screenshot from 2016-04-02 21 09 24


RReverser:
Ah, that's why the highlighting was so bad. Thank you @TimothyGu!

PR #111, opened 2016-04-06
Switch to redux

fkling:
First attempt to switch to redux and redux-sagas. I'm still using PubSub to circumvent the whole system in some cases.

This can certainly be improved. I feel like the reducer is not doing much more than setting values. Maybe more logic should be moved to the reducer.

Parsing the code was moved in `ASTOutput`, i.e. the AST is not part of the global state anymore, which made the sagas and reducer a bit easier.

With this change it will be a lot easier to create other builds for various purposes, such as creating an embeddable version (which I plan to tackle next).

Just putting this up as a PR in case someone wants to comment on it :) Feedback is welcome!


hzoo:
:tada: hope to look at this again soon..

Issue #44, opened 2015-11-03
Let tools specify parser settings


fkling:
Currently tools can only specify which parser to use (so it is selected by default) but not parser settings, which is required as well sometimes.

hzoo:
Ah yeah It would be useful for babel to only run the plugin you are creating (similar to eslint without any rules except the one being created)

RReverser:
@hzoo That's pretty easy to do even without configs - just need to change couple of lines in current code.

hzoo:
@RReverser would we want an option for that or just not transform anything else by default?

RReverser:
@hzoo I believe that should be default as that's usually what user expects (mixing built-in transformation combined with own one is confusing when you're trying to develop a plugin).

hzoo:
@RReverser Ok can do a PR - so we can just remove those deps then?

PR #112, opened 2016-04-20
Only transform the user's plugin

hzoo:
Ref https://github.com/fkling/astexplorer/issues/44

screen shot 2016-04-20 at 9 11 45 pm

screen shot 2016-04-20 at 9 12 33 pm

cc @RReverser

### DIFF
--- a/src/parsers/js/transformers/babel6/index.js
+++ b/src/parsers/js/transformers/babel6/index.js
@@ -26,7 +26,7 @@ export default {
     );
 
     return babel.transform(code, {
-      presets,
+      presets: [],
       plugins: [(transform.default || transform)(babel)],
       sourceMaps: true,
     });
--- a/src/parsers/js/transformers/babel/index.js
+++ b/src/parsers/js/transformers/babel/index.js
@@ -25,7 +25,7 @@ export default {
     );
 
     return babel.transform(code, {
-      ...options,
+      whitelist: [],
       plugins: [transform],
       sourceMaps: true,
     });


src/parsers/js/transformers/babel6/index.js
```diff
@@ -26,7 +26,7 @@ export default {
     );
 
     return babel.transform(code, {
-      presets,
+      presets: [],
```
hzoo: can do this or just remove the key

RReverser: Yeah, I remember it was pretty easy for Babel 6, but dirtier for Babel 5, but perhaps worth including both in this PR? Thanks!

hzoo: Oh what do you mean by both in this case? is it good?

hzoo: Oh... lol I see what you mean now

hzoo: Interesting.. I kept wondering why I couldn't use any stage-0 features in the code and got unexpected token.

It's because of `halts(transformCode);` which uses acorn and doesn't understand stage-x syntax.

hzoo:
Ok done + added a screenshot

src/parsers/js/transformers/babel6/index.js
RReverser: Ah yes - now I recall why I gave up when trying to do this. You need to include all the syntactic plugins manually when you remove transformation presets.

fkling: > Interesting.. I kept wondering why I couldn't use any stage-0 features in the code and got unexpected token.

So, does everything work now or not? Are you saying one cannot use stage 0 features in the transform?

hzoo: Everything works yeah. I added screenshots above to show.

I mentioned you can't use stage 0 features in the transform itself because the halting code (`halts(transformCode);`) we added it gives an unexpected token error (because it uses acorn to parse the code).

hzoo: I think we'll need to modify `halting-problem` to allow using `babylon` instead like in https://github.com/ForbesLindesay/halting-problem/blob/master/lib/halting-problem.js#L3? cc @ForbesLindesay

fkling: Ah OK. I thought something that worked before wouldn't work anymore, but that doesn't seem to have worked, ever.

STRML:
This broke a bunch of code - for instance the snippet for [babel-plugin-flow-react-proptypes](http://astexplorer.net/#/YzN9Yq2vLH/4), which requires the Flow parser plugin to work properly.

RReverser:
Yeah, that's what I said /cc @hzoo when talked about need to include parser plugins separately.

PR #122, opened 2016-05-08
Remove `parent` property from TypeScript AST

nmalaguti:
The `parent` property causes many AST elements to highlight rather than the selected element. Since the information is already visualized as a tree, the `parent` property provides no additional information.

### DIFF
--- a/src/parsers/js/typescript.js
+++ b/src/parsers/js/typescript.js
@@ -101,7 +101,8 @@ export default {
     for (let prop in node) {
       if (
         prop === 'constructor' ||
-        prop.charAt(0) === '_'
+        prop.charAt(0) === '_' ||
+        prop === 'parent'
       ) {
         continue;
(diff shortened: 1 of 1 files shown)


fkling:
I thought it could be useful so that people know that `parent` exists at least? Most other parser don't have that. OTOH, I understand that the highlighting is annoying. I'm wondering if there is something we can do about that (independently of whether we should show `parent` or not).

nmalaguti:
Sorry for the delay. I agree that knowing that a `parent` property exists is useful when working with the AST, but it does make the highlighting a lot more difficult to follow.

I haven't looked into whether there's a good way to exclude the parent from being highlighted while still showing the property.

Alternatively we could add a language option to hide the parent properties and make it user configurable.

fkling:
We'll just do this and see what happens ;)

PR #125, opened 2016-06-06
Update default parser for JSCodeShift transform

DrewML:
I believe this matches [the current parser used by JSCodeShift](https://github.com/facebook/jscodeshift/issues/14)


fkling:
jscodeshift still uses recast. You can select babylon in recast's own options though. Ideally the transform can configure recast to preselect babylon. This is not possible yet, but we will get there.

DrewML:
Thanks for clarifying - I must have misunderstood the linked PR.

fkling:
Yeah, we didn't replace recast with babylon, but we are passing babylon *to* recast :)

Issue #115, opened 2016-04-24
Selected Parser not persisted upon save


DrewML:
To start with, thanks so much for the excellent tool!

Reading https://github.com/fkling/astexplorer/issues/67#issuecomment-162856324, it leads me to believe that the expected behavior currently is that the selected parser should be persisted when you `Save` a transform. That does not seem to be working for me currently with new transforms I create on the site.

Was there a recent regression, or am I possibly just overlooking something?

I should note that, the selected parser does still show for the example that was linked in #67 (http://astexplorer.net/#/np0DfVT78g).

RReverser:
@DrewML The link you provided opens with PostCSS parser selected. Which one would you expect instead?

fkling:
I assume what happens is that you selected a parser that is different from the transform's default parser. If you have a transform, then the transform's default parser will always selected, even if you change the parser. This isn't really documented anywhere, so the confusion is understandable.

If this is indeed the problem, we can discuss if this behavior should be changed. E.g. this behavior would make sense for transformers which can work with multiple parsers.

DrewML:
@RReverser Thanks for looking! That link I provided at the end was just an example showing that a parser setting does persist for _some_ saved transforms. I probably should have provided an example of one where it didn't work as well :)

@fkling You nailed it - that's definitely the scenario. [Here is an example transform](https://astexplorer.net/#/ECeWqibQcR/15).

Originally I was changing the parser itself because I didn't think `recast` played well with the spec for class properties. However, I didn't realize that I could change the parser `recast` uses to `babel-core`. Now that I've found that setting, this issue isn't much of a problem for me, since the `recast` settings _do_ persist upon save

fkling:
> since the recast settings do persist upon save

Only locally though ;) Parser settings are not stored in Parse yet. I guess we should do that to at some point. Right now, the parser settings are only accessible from within each parser. I want to refactor how parser settings work soon, so that they can be passed in from the outside (e.g. an API). Then it would also be easier to save the settings with the code snippet.

DrewML:
Cool. Let me know if there is anything I can do to help/contribute.

fkling:
Parser settings are now saved, so this shouldn't be an issue anymore!

DrewML:
You rock @fkling - thank you!

Issue #55, opened 2015-11-23
Persist parser and parser settings in URL


fkling:
To make sharing code snippets easier, it would be convenient to store the selected parser and parser settings in the URL.

However, I don't really like to update the current URI with that information, since it will look ugly and mess with the browser history. Instead, we could add a "share" button that shows a copyable version of the URL, just like jsfiddle does.

I'm already working on this.

RReverser:
Half-done with introduction of saving parser?

fkling:
Yeah... different from what I intended to do, but sufficient enough :)

PR #127, opened 2016-06-12
Store parser settings in snippets

fkling:
This makes sharing settings more consistent. Instead of using the
locally stored parser settings, the snippet parser settings will be
used.

Note however that any changes to settings after a snippet was loaded
will replace the locally stored settings. This might make settings
behavior a little less predictable.

Also note that it's not clear yet what impact stored settings will have
when the corresponding parser is updated and the available options
change.

This would kind of solve #115 and #55 (instead of storing the settings in the URL).

Making a PR to get some feedback on the general idea.


fkling:
I guess nobody cares too much about this :P

hzoo:
I care 😢, maybe not everyone actually watches repos (you should ping the creators of the issues)

> Note however that any changes to settings after a snippet was loaded
will replace the locally stored settings. 

Maybe need to add a notification somewhere saying that the snippet/url changed the settings

Also note that it's not clear yet what impact stored settings will have
when the corresponding parser is updated and the available options
change.

> So if we removed an option in the parser and it tries to set it, or the option takes a different value and tries to set it?

fkling:
> So if we removed an option in the parser and it tries to set it, or the option takes a different value and tries to set it?

Yep.

---

Maybe I have to rethink the configuration system a little so that it is possible to have "your" (local) settings and have "snippet settings".

fkling:
Did a lot of changes to the internal app state, see commit message. Will try to merge this tomorrow.

fkling:
Since this makes changes to the parse backend but we are moving away from parse, I'm going to merge this into new working branch for moving over gists.

fkling:
This was implemented with 5745dbd127c0d4fdcfd786937aa6096e65982fb7 .

PR #129, opened 2016-06-15
gitignore packages/*/node_modules

hzoo:
Is it just me or should we be ignoring node_modules in `packages/`? for example https://github.com/fkling/astexplorer/tree/master/packages/babel6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /out
-/node_modules
+node_modules


fkling:
Are modules actually stored in that location? I might have setup a global `node_modules` rule, that's why this doesn't affect me. But seems reasonable :)

PR #139, opened 2016-07-13
add esformatter to parsers

millermedeiros:
The AST used by [esformatter](https://github.com/millermedeiros/esformatter) is very similar to `babel-eslint`, but we also have a linked list of tokens (which can be accessed through the `startToken`/`endToken` props); Nodes that are inside Arrays, also have `next` and `prev` properties (see [rocambole](https://github.com/millermedeiros/rocambole) for more details). Given these differences, I believe it's worth the addition.

This will make it easier for contributors to [write plugins](https://github.com/millermedeiros/esformatter/wiki/Plugins) and fix bugs.

screen shot 2016-07-13 at 12 57 54 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "cssom": "^0.3.0",
     "domhandler": "^2.3.0",
     "escodegen": "^1.4.1",
+    "esformatter-parser": "^1.0.0",
     "eslint1": "file:packages/eslint1",
     "eslint2": "file:packages/eslint2",
     "espree": "^3.1.0",
--- a/src/parsers/js/esformatter.js
+++ b/src/parsers/js/esformatter.js
@@ -1,0 +1,60 @@
+import defaultParserInterface from './utils/defaultESTreeParserInterface';
+import pkg from 'esformatter-parser/package.json';
+
+const ID = 'esformatter-parser';
+const name = 'esformatter';
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: name,
+  version: pkg.version,
+  homepage: pkg.homepage,
+  locationProps: new Set(['loc', 'start', 'end', 'range']),
+
+  loadParser(callback) {
+    require(['esformatter-parser'], (parser) => {
+      callback(parser);
+    });
+  },
+
+  parse(parser, code) {
+    return parser.parse(code);
+  },
+
+  *forEachProperty(node) {
+    for (let prop in node) {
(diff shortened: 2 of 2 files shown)


RReverser:
> The AST used by esformatter is very similar to babel-eslint

So... is it ESTree-compliant?

millermedeiros:
@RReverser It's a superset of AST returned by [acorn-to-esprima](https://github.com/babel/acorn-to-esprima/)

RReverser:
Rechecked myself, so it appears that `esformatter-parser` is not really parser itself, but just a wrapper around `babylon` that adds extra properties for convenience of `esformatter`, while generally preserving the same AST.

I have mixed feelings about adding such light wrappers to the ever-growing list of parsers, but that's anyway for @fkling to decide.

TimothyGu:
How about we add support for visualization of an array of tokens? That'd solve the too-many-parsers problem.

RReverser:
@TimothyGu Hm, how are tokens (which are not even standardized and each tool uses its own) related to this issue or AST exploration?

TimothyGu:
The point of the wrapper is to provide `startToken` and other token-related AST node properties. If we provide a way to directly look at the list of tokens, support for such wrappers would not be necessary, I would think.

millermedeiros:
@TimothyGu the tokens returned by `esformatter-parser` includes white spaces, line breaks, comments and indentation. It also handles some edge-cases related to the `Program` range (to include comments/shebang/whitespace).

So even if you create a way to visualize the tokens, it would need to include the proper types (eg. `rocambole` uses `{type: 'LineComment', ...}` and `{type: 'BlockComment', ...}` while other tools might use a single `{type: 'Comment', block: true, ...}`)

If the displayed information doesn't match the actual AST, it would be worse than having no tool to *explore* it at all ("a bad map is worse than no map at all").

If this doesn't get merged to `astexplorer` I will end up forking it and creating a version that works only for `esformatter`... - but I'd rather have it here.

hzoo:
I think it's fine, we could even change out the current acorn-to-esprima for this

TimothyGu:
@millermedeiros, makes sense.

fkling:
I'm OK with adding this. In the long run it might be interesting to have a new concept of a "wrapper" which can be turned on for the corresponding parser.

E.g. if `esformatter` is based on babylon, there would be checkbox for the for babylon that enables this wrapper.

Or maybe this makes everything more complicated... it's just something to think about.

PR #143, opened 2016-08-18
Update babylon to 6.9.0

danez:
Is it intended to also commit records.json? Wasn't sure why it changed


fkling:
> Is it intended to also commit records.json?

I haven't thought about it yet. I'm not even sure whether it's that useful. But lets say yes for the time being.

Thanks!

hzoo:
Nice, there were a good amount of bug fixes (like the exponentiation operator issue)

fkling:
I forgot to update the local packages *facepalm*.

hzoo:
yea although acorn-to-esprima isn't really updated.. I moved the logic back into babel-eslint since it was annoying and no one else was using the package at the time.

fkling:
I see... should we replace it with babel-eslint then? Or remove it? I have to confess that I'm confused about the dependencies between acorn-to-esprima, and eslint1 and eslint2.

hzoo:
acorn-to-esprima is just a separate module I made out of babel-eslint that did what the name was.

However, it was annoying to modify in a separate repo so I just stopped updating that. It is necessary because eslint used espree/esprima. However now espree is actually a wrapper around acorn (which is crazy because babel-eslint is doing the same thing as espree).

babel-eslint: babylon parser -> a lot of hacks + convert expreimental syntax into something else -> convert acorn tokens to esprima -> eslint.

I think the issue was that using babel-eslint had issues or something about the bundle being big I forgot

fkling:
@hzoo: Bundle being big could probably be avoided through wepback configuration.

To me it sounds like we should just try to replace acorn-to-esprima here with the latest version of babel-eslint.

PR #145, opened 2016-09-07
flow-parser: Update to v0.32.0

motiz88:
Flow v0.32.0 introduces new syntax, so a new version of `flow-parser` is out and it would be useful here.


fkling:
Thanks!

Issue #147, opened 2016-09-10
Syntax update needed in the AST Explorer example for PostCSS


alexlibby:
I've been using the ASTExplorer for PostCSS - it's a great tool for building up plugins.

However, I've been made aware of a change to the naming convention for PostCSS plugins, and that the AST Explorer example at http://astexplorer.net/#/np0DfVT78g/1 needs to be updated. I'm not sure who can do this: is anyone able to help please?

The details of the change are that references to ```css``` need to be changed to ```root```:
```
import * as postcss from 'postcss';

export default postcss.plugin('postcss-reverse-props', (options = {}) => {
     // Work with options here
return root => {
         // Transform each rule here
root.walkDecls(decl => {
             // Transform each property declaration here
             decl.prop = decl.prop.split('').reverse().join('');
         });
     };
});
```
I've been in contact with the developer for PostCSS, who is aware of the need to change the code.

fkling:
Anybody can go to http://astexplorer.net/#/np0DfVT78g/1 and create a new revision, but it's **not** possible to change the code and keep the same revision number (i.e. `np0DfVT78g/1`), if that's what you are asking about.

If the problem is that the URL is to be preserved, then there is no solution for this at the moment. Eventually, when we move to gists, then there will be a way to always refer to the latest revision. Until then I recommend to use some kind of URL redirect from a permanent URL to the astexplorer URL.

If that's not an issue, just go the the example, update the code, click save and use the the new URL.

danharper:
I assume this is referring to the default PostCSS transform example; I've opened an update for it #148.

fkling:
So, the default code example was updated in #148 , but http://astexplorer.net/#/np0DfVT78g/1 wasn't. Anybody can create a new version of http://astexplorer.net/#/np0DfVT78g/ to update it but it's not possible to replace the code of a specific version. Every change will create a new version.

PR #148, opened 2016-09-10
Update PostCSS transform example

danharper:
To match preferred code style in transform plugins (guess this is referring to https://github.com/postcss/postcss-plugin-boilerplate/commit/4d05aabcacd5648bb115108cfff1f4831fb190f2)

for #147


alexlibby:
Just to confirm - this is correct: the main developer for PostCSS has updated the plugin boilerplate code, so this change needs to be reflected in the transform example at http://astexplorer.net/#/np0DfVT78g/1

fkling:
Thank you!

PR #150, opened 2016-09-15
build: fix for Node 6

### LINKED ISSUE #137: Cannot build package with Node v6
I'm using node v6.0.0. When I try to build your project clone, get an error: "ERROR in Path must be a string. Received undefined"

thorn0:
The exception `Path must be a string` comes from `html-webpack-plugin`, v1.7.0, [this line](https://github.com/ampedandwired/html-webpack-plugin/blob/v1.7.0/index.js#L205).

Passing `filename` explicitly fixes it.

Closes #137

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -61,6 +61,7 @@ var plugins = [
 
       return fs.readFileSync('./index.html', 'utf-8');
     },
+    filename: 'index.html',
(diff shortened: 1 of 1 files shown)


fkling:
Yay! That is great! I tried upgrading html-wepack-plugin but that didn't work. Thank you!

PR #151, opened 2016-09-19
Update jscodeshift default

vjeux:
A few changes to hopefully make it more beginner friendly

- Rename the variable `p` to `path` to make it more explicit
- Add `{}` inside of the arrow function in order to be able to add `console.log` more easily
- Use `forEach` instead of directly `replaceWith` as it's extremely likely that the codemod the person is going to write is not going to be as straightforward as just replacing a node with another one.
- Remove the default template stuff. I looked for [cpojer/js-codemods](https://github.com/cpojer/js-codemod) & [reactjs/react-codemod](https://github.com/reactjs/react-codemod) and there isn't a single use of those. We should document how they work and get people to use it, but I don't think it is useful to keep them by default.
- Remove the last `;` as it triggers a lint error on cpojer/js-codemods

I believe that this is going to be make the default experience more developer friendly. Let me know what you think :)

cc @fkling

### DIFF
--- a/src/parsers/js/transformers/jscodeshift/codeExample.txt
+++ b/src/parsers/js/transformers/jscodeshift/codeExample.txt
@@ -9,4 +9,4 @@ export default function transformer(file, api) {
       );
     })
     .toSource();
-};
+}
--- a/src/parsers/js/transformers/jscodeshift/codeExample.txt
+++ b/src/parsers/js/transformers/jscodeshift/codeExample.txt
@@ -1,11 +1,12 @@
 export default function transformer(file, api) {
   const j = api.jscodeshift;
-  const {expression, statement, statements} = j.template;
 
   return j(file.source)
     .find(j.Identifier)
-    .replaceWith(
-      p => j.identifier(p.node.name.split('').reverse().join(''))
-    )
+    .forEach(path => {
+      j(path).replaceWith(
+        j.identifier(path.node.name.split('').reverse().join(''))
+      );
+    })
     .toSource();
 };


cpojer:
this is great. @fkling can we merge this? :)

fkling:
> Remove the default template stuff.

OK 😞 

It's super powerful, but you are right, if we don't have proper documentation at this point, it probably isn't that useful.

vjeux:
If you show me how it works I'll document it :)

fkling:
@vjeux: Could be part of your readme2 thingy :)

The template functions allow you to build new nodes more easily, and you can iterpolate existing AST nodes. E.g. instead of

```js
j.memberExpression(identifier,
  j.binaryExpression('-',
    j.memberExpression(identifier, j.identifier('length')),
    j.literal(1)
  )
)
```

you can write

```js
j.template.expression`${identifier}[${identifier}.length - 1]`
```

It has limitations but works very well for expressions.

vjeux:
Yeah, it would fit like a charm inside of the "modify the ast" section.

It's getting better every evening :)
https://github.com/vjeux/jscodeshift/blob/patch-1/README.md

fkling:
This looks really great, thank you for doing this! Note: there is also this PR for adding autogenerated documentation in the `docs/` folder: https://github.com/facebook/jscodeshift/pull/146  . Not sure how we should resolve this.

vjeux:
Template helpers are awesome indeed!

Documented: https://github.com/vjeux/jscodeshift/blob/patch-1/docs/modifying-the-ast.md#template-helpers

fkling:
Great! How should I provide feedback about this or should I wait and comment on the PR?

For example, in the template example, `name` should be an AST node, not a string. The template helpers interpolate AST nodes, not strings.

vjeux:
I just added you as an admin of the repo, feel free to use the github interface to edit the file and commit it there directly :)

vjeux:
By the way, it may not be intended but it does work with a string :)

PR #153, opened 2016-09-19
add offline support

kentcdodds:
This commit updates the `webpack.config.js` file to use the `OfflinePlugin` by @NekR so I can get to and use ASTExplorer.com on the airplane :)

Will require additional changes ([for example](https://github.com/kentcdodds/es6-todomvc/compare/FEM/06.2-extract-css...FEM/07.0-offline)). I'm mostly opening this as a PR rather than an issue to discuss it :)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,6 +3,7 @@ var path = require('path');
 var webpack = require('webpack');
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 var ExtractTextPlugin = require('extract-text-webpack-plugin')
+var OfflinePlugin = require('offline-plugin')
 var ChunkManifestPlugin = require('chunk-manifest-webpack-plugin');
 
 const DEV = process.env.NODE_ENV !== 'production';
@@ -70,6 +71,7 @@ var plugins = [
     process.stdout.write(/build modules/.test(msg) ? '\r' : '\n');
     process.stdout.write(msg);
   }),
+  new OfflinePlugin(),
 ];
 
 module.exports = Object.assign({


NekR:
Just a side note about this: https://github.com/kentcdodds/es6-todomvc/compare/FEM/06.2-extract-css...FEM/07.0-offline#diff-fc1b37867580a192c480780240064424R28

You may just remove `new OfflinePlugin()` from `webpack.config` in a dev mode, if that makes sense. The result will be the same as with your solution.

kentcdodds:
Oh really? What will happen with my call to `offlineInstall`?

NekR:
It will work and write a warning into browser's console that you could
forget to add the plugin to your config. If you care what it will include
in a bundle -- nothing except of a warning.

Here is the content of the runtime by default:
https://github.com/NekR/offline-plugin/blob/master/runtime.js

When you add offline-plugin into the config, it automagically serves real
runtime for you.

On Sep 19, 2016 23:12, "Kent C. Dodds" <notifications@github.com> wrote:

Oh really? What will happen with my call to offlineInstall?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/fkling/astexplorer/pull/153#issuecomment-248110594>,
or mute
the thread

NekR:
Though, I already have reports what that warning is a bit annoying, so I'm
going to so something with it in a next release.

fkling:
This is a great idea! Ever since I heard about the plugin I thought about using it for astexplorer. What else specifically do we have to do?

kentcdodds:
Well, I heard that there's some integration with Parse.com somewhere. But I think for this first iteration it would be enough to just be able to use the tool, even if it means you can't save what you're working on.

So I'm pretty certain that all you'd need to do is add this snippet of code to somewhere in the project (most likely somewhere around the bootstrapping phase of the app):

```javascript
import {install as offlineInstall} from 'offline-plugin/runtime'
offlineInstall()
```

Pretty sure that's it... Maybe further down the road we can disable the "Save" button when you're offline and then add the option to save to localStorage.

gabro:
Hi guys, is there any update on this? It looks promising and I was wondering whether I can help :)

kentcdodds:
No update. I'm sure a more thorough PR would be appreciated. I know I would love it!

PR #154, opened 2016-09-29
update babel, babylon

hzoo:
I removed the `babel-core` from devDeps

http://babeljs.io/blog/2016/09/28/6.16.0#new-feature

Babel will now take in a nested option, I think we have to modify the settings component to support it?

```
{
  "parserOpts": {
    "allowImportExportEverywhere": true,
    "allowReturnOutsideFunction": true,
    "sourceType": "module",
    "plugins": ["flow"]
  }
}
  "parserOpts": {
    "parser": "recast"
  },
  "generatorOpts": {
    "generator": "recast"
  }
```

Then we can write babel codemods with astexplorer as well


fkling:
Sorry that I haven't reacted to this yet.

> Babel will now take in a nested option, I think we have to modify the settings component to support it?

It's currently not even possible to configure transformers.

PR #160, opened 2016-11-03
Use production RawGit URL

captbaritone:
RawGit.com suggests using the `cdn.` subdomain on assets which are not
subject to change. The non-cdn URLs are subject to throttling or rate limiting.

### DIFF
--- a/index.html
+++ b/index.html
@@ -35,5 +35,5 @@
     </script>
     <script async src='https://www.google-analytics.com/analytics.js'></script>
     <!-- End Google Analytics -->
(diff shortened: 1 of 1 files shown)


fkling:
Thanks for that, but I just fixed the actual problem 😃  (will comment in the other task).

captbaritone:
Even better!

fkling:
I mean, maybe it would makes to switch over to rawgit anyway, but that requires some webpack configuration, because the script tags are automatically inserted.

Thanks again!

PR #161, opened 2016-11-04
Use recast in babel transform

hzoo:
@fkling one thing about the yarn.lock change is that the urls changed to `https://registry.yarnpkg.com/acorn-to-esprima` instead of `https://registry.npmjs.org/acorn-to-esprima

Not sure if we want this by default (seems fine?)

Babel can now take in `recast@0.11.16` (`ast-types@0.9.1`) as a parser/generator

```js
import 'recast' from 'recast';
babel.transform(code, {
  parserOpts: {
    parser: recast.parse,
  },
  generatorOpts: {
    generator: recast.print,
  }
}
```

```js
// .babelrc
{
  "parserOpts": {
    "parser": "recast"
  },
  "generatorOpts": {
    "generator": "recast",
  }
}
```

multicursor

a local test with object shorthand/method transform

### DIFF
--- a/src/TransformOutput.js
+++ b/src/TransformOutput.js
@@ -41,7 +41,9 @@ function transform(transformer, transformCode, code) {
     return Promise.resolve(result).then(result => {
       let map = null;
       if (typeof result !== 'string') {
-        map = new SourceMapConsumer(result.map);
+        if (result.map) {
+          map = new SourceMapConsumer(result.map);
+        }
         result = result.code;
       }
       return { result, map };
--- a/src/parsers/js/transformers/babel6/index.js
+++ b/src/parsers/js/transformers/babel6/index.js
@@ -1,5 +1,6 @@
 import compileModule from '../../../utils/compileModule';
 import pkg from 'babel6/babel6-package';
+import recast from 'recast';
 
 const ID = 'babelv6';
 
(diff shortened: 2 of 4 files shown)


src/TransformOutput.js
```diff
@@ -41,7 +41,9 @@ function transform(transformer, transformCode, code) {
     return Promise.resolve(result).then(result => {
       let map = null;
       if (typeof result !== 'string') {
-        map = new SourceMapConsumer(result.map);
+        if (result.map) {
```
hzoo: was getting null for result.map for some reason.. might just be a configuration issue

hzoo: https://github.com/benjamn/recast#source-maps

src/parsers/js/transformers/babel6/index.js
```diff
@@ -24,6 +25,12 @@ export default {
     );
 
     return babel.transform(code, {
+      parserOpts: {
+        parser: recast.parse,
```
hzoo: https://github.com/babel/babel/blob/723ca0eef896835713024017d94b772c2394c80b/packages/babel-core/src/transformation/file/index.js#L413-L445

src/parsers/js/transformers/babel6/index.js
```diff
@@ -24,6 +25,12 @@ export default {
...
+      },
+      generatorOpts: {
+        generator: recast.print,
```
hzoo: https://github.com/babel/babel/blob/723ca0eef896835713024017d94b772c2394c80b/packages/babel-core/src/transformation/file/index.js#L596-L620

hzoo:
🎉 awesome!

fkling:
Thanks for this :)

PR #162, opened 2016-11-05
add a name, remove parameter destructuring [skip ci]

hzoo:
and maybe we should actually use babel.types somehow?


src/parsers/js/transformers/babel6/codeExample.txt
```diff
@@ -1,5 +1,8 @@
-export default function ({types: t}) {
+export default function (babel) {
+  let { types: t } = babel;
```
TimothyGu: Should we use `const` instead of `let` here as it's unlikely the user is going to change any property of `babel`?

Issue #163, opened 2016-11-10
Experimental features in transform


hzoo:
Looks like there currently isn't a way to use any experimental features in the bottom left transform code.

![screen shot 2016-11-10 at 10 26 09 am](https://cloud.githubusercontent.com/assets/588473/20182549/1dd535a4-a730-11e6-9bbf-9b5723542144.png)

Even though `transformJSCode` runs babel with es2015 and stage-1.

There's an unexpected token error which is odd but it seems to be coming from the halting-problem module since it uses acorn. So either halting-problem could use babylon somehow or probably better is we run the toES5 function before and then run `halts(toES5(transformCode))`

cc @fkling, @ForbesLindesay

hzoo:
a PR is https://github.com/fkling/astexplorer/pull/164

PR #164, opened 2016-11-10
Allow stage-0 syntax in transform plugins

hzoo:
Fixes https://github.com/fkling/astexplorer/issues/163

### DIFF
--- a/src/TransformOutput.js
+++ b/src/TransformOutput.js
@@ -17,12 +17,13 @@ function transform(transformer, transformCode, code) {
   }
   // Use Promise.resolve(null) to return all errors as rejected promises
   return transformer._promise.then(([realTransformer, toES5]) => {
+    let es5Code = toES5(transformCode);
     // assert that there are no obvious infinite loops
-    halts(transformCode);
+    halts(es5Code);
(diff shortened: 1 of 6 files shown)


src/parsers/utils/transformJSCode.js
```diff
@@ -1,10 +1,10 @@
 import * as babel from 'babel6';
 import es2015 from 'babel-preset-es2015';
-import stage1 from 'babel-preset-stage-1';
```
hzoo: stage-0 is in package.json so not sure why it was stage-1?

src/TransformOutput.js
```diff
@@ -17,12 +17,13 @@ function transform(transformer, transformCode, code) {
   }
   // Use Promise.resolve(null) to return all errors as rejected promises
   return transformer._promise.then(([realTransformer, toES5]) => {
+    let es5Code = toES5(transformCode);
     // assert that there are no obvious infinite loops
-    halts(transformCode);
```
hzoo: both halts/loopProtect are using acorn

hzoo:
Works for me now (tested with class properties)

PR #169, opened 2016-11-15
log error.stack

hzoo:
screen shot 2016-11-15 at 9 05 26 am

to be able to click on the code


fkling:
Does `console.error(error);` not let you click through the call stack?

hzoo:
Yeah I believe it was just a string at the time I tried this

fkling:
This is what I am getting when I add `new Error('...')` in the transform function:

screen shot 2016-12-29 at 5 31 56 pm

Looks like it works fine as it is?

hzoo:
just going to close for now

PR #172, opened 2016-12-09
add links for missing homepage urls

skratchdot:
I saw that a few of the links were missing for various parsers/transformers.  I went ahead and added fallbacks for the missing homepages.  A better fix would be to submit PR's to the upstream repos so each package.json has a `homepage` value, but this is an okay stopgap IMO.

Here are the packages and the homepages I included:

https://www.npmjs.com/package/cssom - https://github.com/NV/CSSOM
https://www.npmjs.com/package/rework - https://github.com/reworkcss/rework
https://www.npmjs.com/package/htmlparser2 - https://github.com/fb55/htmlparser2
https://www.npmjs.com/package/flow - https://flowtype.org/
https://www.npmjs.com/package/jscodeshift - https://github.com/facebook/jscodeshift
https://www.npmjs.com/package/webidl2 - https://github.com/w3c/webidl2.js

Thanks for this great project @fkling !

### DIFF
--- a/src/parsers/css/cssom.js
+++ b/src/parsers/css/cssom.js
@@ -9,7 +9,7 @@ export default {
   id: ID,
   displayName: ID,
   version: pkg.version,
-  homepage: pkg.homepage,
+  homepage: pkg.homepage || 'https://github.com/NV/CSSOM',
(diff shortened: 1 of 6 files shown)


fkling:
Great, thank you!

PR #173, opened 2016-12-09
fixing eslint errors

skratchdot:
I made `JSON.propTypes.ast` match `Tree.propTypes.ast`, see:

https://github.com/fkling/astexplorer/blob/b5c36c1689e3638bc6dddf61f56d0e3f4770adaa/src/components/visualization/Tree.js#L86-L89

I also confirmed my local build works.


fkling:
Thank you 😃

PR #174, opened 2016-12-09
moving toolbar buttons to components/buttons

skratchdot:
Feel free to close/ignore this PR.  It's a very small refactor where I put all the toolbar buttons in their own folder.  There were already 3 buttons that existed: `CategoryButton`, `ParserButton`, `TransformButton`. I moved the 2 other `<div />` tags into their own components: `SaveButton` and `ForkButton`;


fkling:
Nope, that makes sense to me. The structure basically grew organically, so there is certainly need for cleaning some things up :) Thank you!

PR #177, opened 2016-12-10
adding custom graphql logo

skratchdot:
Not sure if it's worth the overhead, but eventually it might come in handy to be able to add custom logos to the app.  I'm not sure if I found the best solution or not with `fontcustom`.

Here's what the app looks like after this PR:

image

I added a README to the `/fontcustom` folder.  I can move to a better location if you have suggestions.  See: https://github.com/skratchdot/astexplorer/tree/graphql-logo/fontcustom#custom-fonts


fkling:
Yeah, quite a lot of code for a tiny logo, but it's OK ;)

skratchdot:
@fkling - agreed!  I felt kinda dirty adding all that.  I've been using `material-ui` lately, and I really like just being able to use SVG's as icons directly in the app.  I'd like to come up w/ a better logo/icon solution in general.  It's been a long time since I've used font-awesome, and it's not so easy to extend, so this definitely needs cleaned up at some point.  I didn't like adding the whole extra folder either, so am definitely all for cleaning this up.

PR #185, opened 2017-01-16
Update "glimmer-syntax" to "@glimmer/syntax" v0.21.0

Turbo87:
The [glimmer](https://github.com/tildeio/glimmer) repo recently changed to publishing to an NPM org instead of several top-level packages, so this PR will replace `glimmer-syntax` with `@glimmer/syntax` and update it to the most current version.

/cc @fkling @mmun

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "webpack": "^1.12.8"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


fkling:
Thank you!

Issue #184, opened 2017-01-14
Progress: Migration from Parse to Gists


fkling:
This issue will outline the steps I am going to take to move away from Parse.

### Context
Parse is discontinuing its service on January 28th. As explained in #142, I am planning to use gists as storage instead. The linked issue explains the advantages that gists have.

### astexplorer server
Using the GitHub API requires to send authentication tokens (either received via OAuth or otherwise) or similar sensitive information. For me, the easiest way to get this to work is to run a server for astexplorer. Having our own server has other advantages as well. For example, we could add support for parsers that are not implemented in JavaScript.  
I will be paying for the server myself at first and will look for other financing options in the future.

### How will code snippets be saved as gists?
At the beginning, all snippets will be saved as *secret* gists owned by the *astexplorer* user, that I created specifically for this. A gist will have up to three files:

- `astexplorer.json`: Containing configuration and settings, such as the parser that is used.
- `code.js`: The code that is parsed.
- `transform.js`: If a transform/tool is selected, this file contains the code of the transform.

Example: https://gist.github.com/astexplorer/28ae2bb52426fa497264674356d0341a

A snippet can then be accessed via a URL similar to
```
https://astexplorer.net/#/gist/28ae2bb52426fa497264674356d0341a/407c5e81320b3bf6a07117e191d4a45936d1f772
```

The first part is the hash of the gist, the second part the hash of the revision. I know the URLs are long and ugly. Maybe we can eventually offer to create short URLs to specific revisions.

**Note:** Everybody that has a URL to such as snippet will be able to add a new revision, just like it works now.

### Can I save the snippets to my GitHub account?
Eventually yes. The first priority is to ensure that astexplorer continues to work. Being able to authenticate will be added later.
The advantage of saving snippets to your account is that only you can add new revisions.

### What happens to the existing code snippets?
Existing URLs and snippets will continue to work. After considering various options, I think the easiest solution is to export the existing data and simply store the exported files on the server and serve them from there.  
Maybe the snippets will be migrated to gists eventually, but that's not a priority.

**Note:** Existing snippets will be loaded "read-only". Changes can only be made by forking the snippet (in which case it will be saved as gist). The UI will point that out.

#### Phase 1
- [x] Deploy server with gist API
- [x] Deploy new site that loads existing snippets read only and stores new snippets in gists

#### Phase 2
- [x] Export final Parse dataset
- [x] Deploy new server version that can serve parse snippets from the  exported files
[...]

PR #186, opened 2017-01-18
Add CSSTree to parsers

lahmatiy:
Added CSSTree – detailed CSS parser (https://github.com/csstree/csstree)

### DIFF
--- a/src/parsers/css/csstree.js
+++ b/src/parsers/css/csstree.js
@@ -17,9 +17,9 @@ export default {
   },
 
   parse(csstree, code) {
-    return JSON.parse(JSON.stringify(csstree.parse(code, {
+    return csstree.toPlainObject(csstree.parse(code, {
       positions: true,
-    })));
+    }));
   },
 
   getNodeName(node) {
--- a/src/parsers/css/csstree.js
+++ b/src/parsers/css/csstree.js
@@ -18,7 +18,7 @@ export default {
 
   parse(csstree, code) {
     return JSON.parse(JSON.stringify(csstree.parse(code, {
-      positions: true
+      positions: true,
     })));
   },
 
@@ -34,5 +34,5 @@ export default {
 
   opensByDefault(node, key) {
     return key === 'children';
-  }
+  },
 };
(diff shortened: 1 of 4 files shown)


src/parsers/css/csstree.js
```diff
@@ -0,0 +1,38 @@
...
+  version: pkg.version,
+  homepage: pkg.homepage || 'https://github.com/csstree/csstree',
+  locationProps: new Set(['loc']),
+
+  loadParser(callback) {
+    require(['css-tree'], callback);
+  },
+
+  parse(csstree, code) {
+    return JSON.parse(JSON.stringify(csstree.parse(code, {
```
fkling: Why is this needed here? Does `.parse` not return a plain object?

lahmatiy: It does, but CSSTree AST stores children nodes as a bi-directional list (a List instance), not as an array. If it were possible to make preprocessing of values, it would be easier:

```js
preprocessValue(value) {
    return value instanceof csstree.List ? value.toArray() : value;
}
```

Or we can convert `List` instances to array by walker:

```js
parse(csstree, code) {
    var ast = csstree.parse(code, {
        positions: true
    });
    csstree.walkUp(ast, function(node) {
        if (node.children && node.children instanceof csstree.List) {
            node.children = node.children.toArray();
        }
    });
    return ast;
}
```
It's more verbose, but if you prefer that way I can update the PR ;)
Maybe there will be a method to make it easier out of the box. But I'm still thinking about it.

lahmatiy: Updated `csstree` to new version with `toPlainObject()` method on board. Now it looks better ;)

lahmatiy:
@fkling Is there a chance for merge?

fkling:
Yeah, sorry for the delay. I'm currently focusing on a big change ( #184 ). I will merge your changes on top of that and publish a new version either today or tomorrow.

lahmatiy:
@fkling Cool! Thank you!

fkling:
Sorry, I didn't find much time to work on astexplorer the last two days. I'll definitely get to it this week though.

lahmatiy:
It's ok. Thank you

fkling:
I rebased and merged this PR in 87bc4f879b8882399517a09420ad5cd9f63ffabf . Thank you!

lahmatiy:
Awesome! Thank you so much 👍

PR #187, opened 2017-01-19
Add wireformat view for glimmer transform

Turbo87:
screenshot

This PR also introduces the possibility to return a POJO from the transformer and have it displayed inside the `JSONEditor` component.

/cc @fkling @mmun

### DIFF
--- a/website/css/style.css
+++ b/website/css/style.css
@@ -314,6 +314,10 @@ html {
   bottom: 0;
 }
 
+.output > .no-toolbar {
+  top: 0;
+}
+
 #JSONEditor .CodeMirror {
   font-size: 0.9em;
 }
--- a/website/package.json
+++ b/website/package.json
@@ -35,6 +35,7 @@
     "webpack": "^1.12.8"
   },
   "dependencies": {
+    "@glimmer/compiler": "^0.21.2",
     "@glimmer/syntax": "^0.21.2",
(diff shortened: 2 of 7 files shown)


fkling:
Sorry for the late reply. I like the idea of returning an object from the transformer and showing it in the JSON view! This could be useful for other transformers too!

I was wondering whether it is necessary to add a new transformer though. I think it would be more convenient if the existing transform could handle both cases. However, it looks like that would require being able to configure the transformer somehow, which is currently not possible, but which is something that has come up before.

If this PR can wait a little I could take a stab at that after this week (after #184 is done).

Turbo87:
> I was wondering whether it is necessary to add a new transformer though. I think it would be more convenient if the existing transform could handle both cases. However, it looks like that would require being able to configure the transformer somehow, which is currently not possible, but which is something that has come up before.

yes, that would've certainly been a better way.

> If this PR can wait a little

I'm certainly not in a rush, but you could also merge this and then you'd have something to refactor once you get around to implementing transformer configs. I'm fine either way. Thanks again for working on this awesome tool! :)

Turbo87:
@fkling rebased to fix conflicts

fkling:
👍 

I think we'll be able to consolidate both transforms in the future without being to hacky.

mmun:
Awesome!

PR #190, opened 2017-01-27
Upgrade to flow-parser@0.38.0

mroch:
the yarn.lock has a few more changes than I initially expected. it looks like because flow-parser@0.37.0 required ast-types@0.18.15, yarn chose to install recast@0.10.43 which also uses that dependency. now, flow-parser@0.38.0 requires ast-types@0.18.18 and yarn thinks recast is better off using the older 0.10.33 that depends on ast-types@0.18.12 which it already has. there's probably no version of recast that depends on ast-types@0.18.18.

I ran this with yarn 0.19.1

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -62,7 +62,7 @@
     "eslint2": "file:packages/eslint2",
     "espree": "^3.1.0",
     "esprima": "^3.1.3",
-    "flow-parser": "^0.37.0",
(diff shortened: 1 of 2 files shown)


fkling:
Thank you! I will merge this later at home.

One day we will have astexplorer upgrade the packages automatically.

PR #191, opened 2017-01-29
fix path to ../out directory

skratchdot:
Now that the folder structure has been re-organized, I tried:
```
cd website;
yarn start;
```
Which no longer served the website. This fixes the path to the `out` folder.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -97,7 +97,7 @@
     "yaml-ast-parser": "^0.0.31"
   },
   "scripts": {
-    "start": "http-server out",
+    "start": "http-server ../out",
     "build": "rimraf ../out/* && cross-env NODE_ENV=production webpack",
     "watch": "webpack -dw",
     "push": "./scripts/push-site.sh",


Turbo87:
this bit me too. but I'm wondering if instead the `out` folder should actually be moved inside of the `website` folder?

skratchdot:
Yeah. I don't really have a preference.  If we pull this in, or move `out` to `website`, there's still some more cleanup that needs to be done IMO:
- `./scripts/push-site.sh` no longer exists, so the `push` task needs removed (or rewritten)
- NICE TO HAVE: have `npm/yarn start` start a `webpack-dev-server` instead of using `http-server` (so we can get live-reloading/etc).

fkling:
yeah, it's not necessary anymore to have the `out/` folder in the top level directory. I kept it for push safety so that GitHub pages continues to work, but by now the new DNS records should have propagated (i.e. GitHub pages isn't used anymore).

fkling:
Forget what I said before... `out/` was only used for copying the build artifacts to the `gh-branch`.

PR #203, opened 2017-03-03
Use postcss-loader + Update dependencies

danez:
I initially only wanted to update babylon to 6.16.0, but seems there is no way unless modifying package.json in babel6/. So I did `yarn upgrade` which also updated a lot of other dependencies to their latest version. If that is not okay I can also only try to update the parsers themself (babylon6, espree, acorn, recast, ...)

While testing I also noticed that webpack is throwing deprecated warnings, so I updated `extract-text-webpack-plugin` to 2.0 instead of beta and fixed the configuration. `autoprefixer-loader`is also deprecated and I instead used postcss-loader with autoprefixer.

I know these are a lot of changes, but hope it is okay.


fkling:
Thanks for doing this! Hopefully rather sooner than later #197 will land and we don't have to worry about upgrading parsers anymore 😉 

I'm going to resolve the merge conflict locally.

fkling:
Merged as 538aa8f43601fe7a4090cfc96f46ac0f96d48706 .

PR #207, opened 2017-03-17
update babel packages

hzoo:
is this all we have to do? forgot


fkling:
That's all :) Thank you!

PR #209, opened 2017-03-20
Update defs

clentfort:
Correct two typos in `scripts/defs/jsdoceshift.yaml` and in the generated `website/src/defs/jscodeshift.json`.  I could not figure out how to run `definition-generator.js`, so I had to fix the errors in the json by hand. :/


fkling:
Thank you!

PR #213, opened 2017-03-27
Add prettier to JSCodeshiftEditor

lewisf:
This allows prettier to run on the JSCodeshiftEditor. I'm not sure if 'blur' is the right time to run this honestly but at the very least I'm wondering if it's worth adding prettier to astexplorer.net somehow (maybe as a togglable option?). I definitely currently spend a lot of time just formatting code within the in-browser editor.

### DIFF
--- a/website/css/style.css
+++ b/website/css/style.css
@@ -487,3 +487,17 @@ body .CodeMirror-Tern-tooltip {
   padding: 5px;
   width: calc(100% - 10px);
 }
+
+.toggleBtn {
+  position: absolute;
+  right: 0;
+  height: 20px;
+  z-index: 10;
+  cursor: pointer;
+  outline: none;
+}
+
+.toggleBtn > .btnText {
+  padding-left: 5px;
+  font-size: 12px;
+}
--- a/website/package.json
+++ b/website/package.json
@@ -85,6 +85,7 @@
     "postcss-less": "^0.15.0",
     "postcss-safe-parser": "^2.0.0",
     "postcss-scss": "^0.4.1",
+    "prettier": "^0.22.0",
     "pubsub-js": "^1.4.2",
     "react": "^15.3.1",
     "react-dom": "^15.3.1",
--- a/website/src/app.js
+++ b/website/src/app.js
@@ -3,7 +3,7 @@ import ASTOutputContainer from './containers/ASTOutputContainer';
 import CodeEditorContainer from './containers/CodeEditorContainer';
 import ErrorMessageContainer from './containers/ErrorMessageContainer';
 import GistBanner from './components/GistBanner';
-import LoadingIndictorContainer from './containers/LoadingIndicatorContainer';
+import LoadingIndicatorContainer from './containers/LoadingIndicatorContainer';
 import PasteDropTargetContainer from './containers/PasteDropTargetContainer';
 import PubSub from 'pubsub-js';
 import React from 'react';
@@ -36,7 +36,7 @@ function App(props) {
       <ErrorMessageContainer />
       <div className={'dropTarget' + (props.hasError ? ' hasError' : '')}>
         <PasteDropTargetContainer>
-        <LoadingIndictorContainer />
+        <LoadingIndicatorContainer />
         <SettingsDialogContainer />
         <ShareDialogContainer />
         <div id="root">
--- a/website/src/components/Editor.js
+++ b/website/src/components/Editor.js
@@ -1,6 +1,17 @@
 import CodeMirror from 'codemirror';
 import PubSub from 'pubsub-js';
 import React from 'react';
+import prettier from 'prettier';
+
+const defaultPrettierOptions = {
+  printWidth: 80,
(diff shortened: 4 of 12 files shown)


fkling:
I can definitely see value in having an option for it, but I don't think enabling this behavior by default is a good idea. I could imagine there quite a lot of people that don't use prettier and want to preserve their own style.

Is this primarily for making the code more easily readable in the editor itself?

lewisf:
Yeah, I think its super helpful for that. I find that my co-workers and write and share codemods we're working on within it and it'd be nice to have the code be more readable as a toggleable option.

lewisf:
@fkling what would seem like a sensible implementation for this?

is there room for this to be a hidden feature that is toggleable by some runtime variable that could be set in the JS console? or would you prefer coming up with with a spot in the UI where this should exist?

pksjce:
Hey I'd love for the transform editor to have formatting with prettier. I routinely write all my mods there and copy it to my files.
I would implement a toggle button at the top right corner of the transform box. @lewisf - I can help you review and complete this if you are available?

fkling:
I have two suggestions:

- Make `prettier` available inside the transform, so if someone wants to use it they could do `return pretter.format(....)` OR
- Similar to @pksjce's suggestion, but instead of adding the button to the transform editor, add it over the *output editor*.

lewisf:
@fkling i think we'd want it to format in the transform editor as you type so that editing a codemod within the transform editor will maintain prettier's formatting.

it doesn't look like there would be as much value from applying the prettier transform on the output, since if someone wanted to use that output, they could copy it into their project and run prettier on their own.

fkling:
Oh I somehow forgot that this was about the transform code itself. Sorry.

lewisf:
@pksjce I don't have time immediately to continue this work -- if you want to take it over that'd be great! if not, I can try to get to it this weekend or next weekend.

pksjce:
@lewisf - I can pick this up. How'd you want to do it? Can you add me as a collaborator or something?

lewisf:
@pksjce just sent you collab rights on my fork, thanks!

pksjce:
@lewisf - Thanks!
@fkling - Hey. I have one question. Why is there one `JSCodeshiftEditor` and an `Editor` component? In this solution, the `Transformer` for `babel` etc will not get the `prettier` support.

fkling:
`JSCodeshiftEditor` extends `Editor`. It loads in a whole bunch of tern stuff for code completion that isn't necessary for other tools.

pksjce:
So, shouldn't I add this `prettier` option to `Editor` itself?

pksjce:
Hey this works now. Would love a review from you!

screen shot 2017-05-13 at 9 01 12 pm

lewisf:
Oh, excited for this. Thanks for working on it @pksjce!

pksjce:
Hey! @fkling - Would love a review on this. I have completed it from my end and have provided the screenshot above :)

fkling:
Looks pretty good, thank you! I was wondering if you think whether there is a need for being able to prettify the "input" code itself (upper left editor) or not? Because if yes, then we could move the button into the `Editor` component itself so that it becomes available to all `Editor` instances.

Thoughts?

pksjce:
Well, the code on top left and bottom right is kind of third party code right, which is not actively edited by the user. At least I don't actively edit it. I simply paste example type of code.
Only this transform editor is where I actually write many lines of code and without some auto formatting I struggle a lot to match all the brackets and parenthesis. This is the pain point I want to solve. Though if someone else from the community seconds you, I'm open to moving the button into the main toolbar.

kentcdodds:
Might be even better if jscodeshift added support for prettier. I would hate to have someone think that they got a really nice codemod written up only to find that when they ran it on their code it messed things up...

kentcdodds:
Either way though, having a toggle could indicate that prettier is unavailable normally...

pksjce:
@kentcdodds- this only allows you to format the codemod editor. It doesn't get involved with the source or transformed code. Those parts still use whichever printer jscodeshift or babel uses

fkling:
@kentcdodds People can already use prettier in their own codemods. They just need to require it and `return prettier.format(whatever);`. But I agree that jscodeshift should support a way to customize printing that is independent of the codemod itself.

fkling:
Lets go ahead with this and update if necessary. Thank you!

fkling:
After trying it I'm thinking maybe it should just be a button that prettifies the code on click. The blur behavior is kind of obscure and doesn't really improve UX over just clicking a button?

pksjce:
Oh you merged it! Actually I was thinking the same thing today. There is no format on save for that editor. But blur is no different than formatting it on button click no? Still its nicer to be explicit. I can raise another PR to do it?

pksjce:
On thinking more, how about if we apply prettier to the mod when return button is pressed?
But would it displace user's cursor position? I'll try this out today and tell you.

Issue #117, opened 2016-04-29
Issue with ESLint v2 or v1` transform + sourceType: module


DrewML:
Hoping this isn't just an obvious case of user-error on my part.

Can be seen at: http://astexplorer.net/#/DH4gO9Bvjr

![image](https://cloud.githubusercontent.com/assets/5233399/14906013/6f9b36a2-0d7c-11e6-8d0c-82a9d924f53b.png)

![image](https://cloud.githubusercontent.com/assets/5233399/14906019/78d245d0-0d7c-11e6-82b6-8277579061a6.png)

fkling:
I have a hunch what the issue is: The transform (eslint) and the parser (espree) are actually completely separate. The parser settings really only affect the parser, which is used to show the tree in the top right panel. ESLint uses its own instance of the parser, with its own settings.

I agree with you that the parser settings should also be used for the transform tool. What prevents us from doing that right now is

- Parser settings are pretty much internal to the parser module itself. Even though we write the settings of every parser to local storage, I have to take a closer look if transform tools could simply read them from local storage without much post-processing (meaning that the settings might stored in a different format than what the transform tool needs).

- Even though we switch to the preferred parser when a transform tool is selected, one is free to choose another one afterwards (e.g. you could switch to traceur). This isn't necessarily a problem, since eslint could simply always read the espree parser settings from local storage, but that might also be surprising. Maybe we should just prevent people from switching parsers when a tool is selected to avoid confusion.


The way parser settings are handled is not ideal at the moment. I want to merge #111 soon and then work on improving parser settings (e.g. change them so that they are passed *to* the parser, which would also make it easier to pass them to the transform tool).

DrewML:
Cool - that makes sense. Thanks for clarifying.

kentcdodds:
[This](https://astexplorer.net/#/gist/c0944762e295413656246dd8c0f60d1c/36a0beb4a7adc7f44aa3673e83736f8e47ee50b7) appears to have come back or still be a problem. How can I help?

PR #219, opened 2017-04-19
improve defaults for eslint transform

kentcdodds:
Ref #117 
I haven't actually tested this out but it seems like this would be the change to make until we add a UI to configure the transform...


kentcdodds:
Hmmm... Tried locally and this doesn't work. I'm not sure what's up :-(

fkling:
Thanks for working on this! I'm currently a bit busy, but will take a closer look next week.

fkling:
So, the problem is with the `parseNoPatch` method actually. That results in an error. The eslint setup is a bit weird, I will submit a new PR that fixes that issue and cleans eslint up.

Your changes are still needed though, so I will merge this. Thank you!

kentcdodds:
Wahoo! Thank you for working on this!

PR #224, opened 2017-05-02
update to TypeScript 2.3

mohsen1:


RReverser:
@fkling Would it perhaps make sense to turn on Greenkeeper for this project so that such PRs would come automatically on each update?

mohsen1:
That would be nice!

fkling:
I'll look into that tonight!

fkling:
So, I enabled greenkeeper, but it might not actually work if the package.json file is not in the root? Or everything is up-to-date 😛

RReverser:
@fkling Huh, apparently not :( https://github.com/greenkeeperio/greenkeeper/issues/139

Any reason not to have it and other configs at the root though?

fkling:
Mmh... I guess not. I just thought it's cleaner this way.

RReverser:
Probably... although there are few config files in the root already, shouldn't hurt too much :P

RReverser:
Oh I see, there are separate package.json's for different folders, so this is indeed the monorepo case.

RReverser:
Then there are only two ways: create org for astexplorer (and add us all :D) with separate repos per project (server/website/...) or continue updating things manually for now...

PR #227, opened 2017-05-17
[eslint] Clean up eslint and babel-eslint

fkling:
The eslint setup in astexplorer is a bit chaotic. I'm not sure if it's worth investing a lot of time to clean this up in the light of  #197 (which I will hopefully finish some day 😛 ), so I just did some minimal changes:

- Switch the default parser for eslint2 and eslint3 to `babel-eslint`, because that's what `eslintUtils` uses.
- Enable `sourceType: 'module'` in `eslintUtils`.
- Make the `babel-eslint` parser actually refer to `babel-eslint` and not `acorn-to-esprima`.
- Since `acorn-to-esprima` is deprecated, hide in from the menu.

In order to really clean this up, `eslint1` would have to use an older version of `babel-eslint` that uses `acorn-to-esprima`, but I didn't want to spend more time with this.

@hzoo , does this look good?

cc @kentcdodds (follow up to #219 )

### DIFF
--- a/website/src/components/buttons/ParserButton.js
+++ b/website/src/components/buttons/ParserButton.js
@@ -13,6 +13,7 @@ export default class ParserButton extends React.Component {
   }
 
   render() {
+    const parsers = this.props.category.parsers.filter(p => p.showInMenu);
     return (
       <div className="button menuButton">
         <span>
@@ -20,7 +21,7 @@ export default class ParserButton extends React.Component {
           &nbsp;{this.props.parser.displayName}
         </span>
         <ul>
-          {this.props.category.parsers.map(parser => (
+          {parsers.map(parser => (
             <li key={parser.id} onClick={this._onClick} data-id={parser.id}>
               <button type="button" >
                 {parser.displayName}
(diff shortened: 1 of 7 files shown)


hzoo:
I think it'd be ok to drop eslint1 anyway since there's an v4 coming out?

kentcdodds:
Honestly I don't see much reason to keep anything but the latest version 🤷‍♂️

fkling:
@kentcdodds Are you saying all my efforts in #197 are for nothing? 😛  jk

> Honestly I don't see much reason to keep anything but the latest version

One reason is to be able to load snippets that were created with an older version of a tool or parser.

> I think it'd be ok to drop eslint1 anyway since there's an v4 coming out?

I thought the same, however, google analytics says that there is still traffic for eslint1 (last 30 days):

screen shot 2017-05-17 at 7 54 46 am

And I believe these are really instances where eslint1 is selected, not just loaded for an existing snippet (but I have to check the code for that 😉 ).

However, in general I think we should only expose the latest version of a tool, but make it still loadable if needed, just like in this case for `acorn-to-esprima`. We kind of get that with #197 anyway.

fkling:
so, good to go? 😃

PR #231, opened 2017-05-25
babel 7

hzoo:
Copied everything from babel5/6 files to make this run (although going to be updating this package often)

screen shot 2017-05-25 at 3 37 24 pm


fkling:
I really should continue with #197 , that would make everything easier...

PR #239, opened 2017-06-14
add GLSL language

gre:
using the glsl-parser library, i had to workaround to get the location but it appears to be correct after workaround (the "end" is not provided by the parser, i guess it should be fixed in the parser itself later https://github.com/stackgl/glsl-parser/issues/17 )

i was not able to gets the syntax color highlighting but i guess it's ok for now as even CSS don't have it.

### DIFF
--- a/website/src/parsers/glsl/glsl-parser.js
+++ b/website/src/parsers/glsl/glsl-parser.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'glsl-parser/package.json';
+import SettingsRenderer from '../utils/SettingsRenderer';
+
+const ID = 'glsl-parser';
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: ID,
+  version: pkg.version,
+  homepage: pkg.homepage,
+  locationProps: new Set(['loc']),
+  _ignoredProperties: new Set(['loc']), // we ignore the loc itself because it's actually a locally enhanced (not in the actual parser data)
+
+  loadParser(callback) {
+    require(['glsl-tokenizer/string', 'glsl-parser/direct'], (
+      tokenize,
+      parse
+    ) => {
+      callback({ tokenize, parse });
+    });
+  },
+
+  parse({ tokenize, parse }, code) {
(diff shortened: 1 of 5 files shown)


website/src/parsers/glsl/glsl-parser.js
```diff
@@ -0,0 +1,71 @@
...
+
+  getNodeName(node) {
+    return node.type;
+  },
+
+  opensByDefault(node, key) {
+    return key === 'children' && node.type === '(program)';
+  },
+
+  renderSettings(parserSettings, onChange) {
```
fkling: There is no need to add this method if there are no settings. In fact, I think it should not be there so that the button is disabled in the UI.

fkling:
Thank you, that's pretty cool!

fkling:
I merged the PR locally to play with it a little and noticed a couple of things:

- There are two "recursive" properties, `stage` and `parent`. While astexplorer has no problem rendering them, it is a bit more usable if such properties are hidden. Not sure if you want that though, it's just a suggestion.
- Some highlights seem off, e.g. this `functionargs` property. Not sure whether that's because of your custom changes or because of the parser, or maybe that's how it is supposed to be.
 screen shot 2017-06-14 at 10 27 01 pm
- What's with all the `placeholder` children? (I'm just curious)

gre:
- you are right, it's pointless to display stage and parent.
- agreed that the functionargs looks wrong. actually it looks like a parser bug to me (see screenshot attached **(1)** , the '(' token looks totally gone which breaks my way of guessing the "end" position for the previous "keyword" node).
- placeholder looks weird yeah, i don't really get it. https://github.com/stackgl/glsl-parser/issues/9


**(1)**

screen shot 2017-06-15 at 14 07 58

gre:
glsl-parser still have bunch of tiny bugs and polish to be done, and unfortunately have not been active for a while but i'll try to make these happen. (i'm part of stackgl org too)

fkling:
Thanks for the updates and sorry for the slow turnaround time. Our family is currently visiting which doesn't leave much time for other things...

fkling:
Merged as 6ed63bf762927e3339968ee2552a6f947efb0591 (I squashed the commits)

Issue #218, opened 2017-04-17
Add typescript-eslint-parser for javascript


unlight:
https://github.com/eslint/typescript-eslint-parser
It's kind of babel-eslint

fkling:
Not sure if I can get around to that, but feel free open a PR! Thanks!

Issue #69, opened 2015-12-08
Switch to CloudFlare?


RReverser:
@fkling Would you be interested in putting astexplorer.net behind CloudFlare? Free plan should work perfectly, and will give HTTPS, HTTP/2 + SPDY and CDN out of the box which should speed up loading our bundles for users.

fkling:
Just looked a bit into it, looks reasonable. Will give it a try soon!

RReverser:
@fkling Let me know should you have any questions :wink:

fkling:
Ah right, you work there :smile:

RReverser:
Guilty :smile: But this is definitely not something I'm paid for, just really think it's a good idea :)

fkling:
No worries, I trust you :) I just registered and enabled it... that was really easy! Now I just need to wait for the nameserver to update.

Given that the JS files are now cached, I think we have to configure webpack to add hashes to the bundle files, and somehow update `index.html` to refer to the correct files. I guess https://www.npmjs.com/package/html-webpack-plugin will help.

RReverser:
@fkling Just so that we don't wait later - did you enable HTTPS, HSTS and HTTP/2?

fkling:
Seem to be enabled by default, except HSTS. Is it worth enabling for us though?

fkling:
Nameservers are now active btw. Can't really test it though since the DNS entries seem to be cached somewhere between my computer and the nameserver ;)

RReverser:
Ah cool. I don't know, maybe not, but in practice way too many Wi-Fis at cafes tend to insert own banners or tracking codes when website is sent over HTTP connection, so better to have HTTPSed everything possible.

RReverser:
Hmm, my nslookup already returns CloudFlare IPs after DNS flush, but seems to be still sent over HTTP.

RReverser:
Ah right I was still visiting HTTP version and HTTP/2 works only over secure connections (one more reason to enforce it ;) ). It works!

RReverser:
Ah right, we don't benefit that much because our bundles are not loaded in parallel and `parse-1.3.0.min.js` is loaded from 3rd-party so blocks most of the page.

fkling:
So this was all for nothing? :P

RReverser:
Fast and secure! :P
If serious - put `parse-1.3.0` onto same server, it should speed it up a bit.
If loading it asynchronously, it would speed up even more, as it's currently the main blocker:

[...]

PR #242, opened 2017-06-18
Add typescript-eslint-parser

josephfrazier:
I adapted this from the `babel-eslint` parser, as described in
https://github.com/fkling/astexplorer/tree/b7ae2f06322c080313fdbea13501a19750d741ca#how-to-add-a-new-parser

This fixes https://github.com/fkling/astexplorer/issues/218.


website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,41 @@
...
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: name,
+  version: pkg.version,
+  homepage: pkg.homepage,
+  locationProps: new Set(['loc', 'start', 'end', 'range']),
+
+  loadParser(callback) {
+    require(['typescript-eslint-parser'], typescriptEslint => callback(typescriptEslint));
```
RReverser: Nit: no need to create extra arrow function here, you can use just `require(['typescript-eslint-parser'], callback)` as we do in many other places.

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,41 @@
+import defaultParserInterface from './utils/defaultESTreeParserInterface';
+import pkg from 'typescript-eslint-parser/package.json';
+
+const ID = 'typescript-eslint-parser';
+const name = 'typescript-eslint-parser';
```
RReverser: You can just use `displayName: ID`, it's fine when they match (we do this in other parsers too).

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,41 @@
...
+      sourceType: 'module',
+    };
+
+    const ast = parser.parse(code, opts);
+    delete ast.tokens;
+    return ast;
+  },
+
+  nodeToRange(node) {
+    if (typeof node.range !== 'undefined') {
```
RReverser: Perhaps just `if (node.range)` would be better since it theoretically can be(come) `null`.

josephfrazier:
Thanks for the feedback, @RReverser! I just pushed up some fixes.

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,40 @@
...
+    require(['typescript-eslint-parser'], callback);
+  },
+
+  parse(parser, code) {
+    const opts = {
+      sourceType: 'module',
+    };
+
+    const ast = parser.parse(code, opts);
+    delete ast.tokens;
```
RReverser: Does `typescript-eslint-parser` return `.tokens`? As far as I see, it does only when you pass explicit option to do so, so there's no point in deleting this property, you can just return result as-is.

RReverser:
Thanks! Added one more comment and - can you please add settings for options that make sense in `typescript-eslint-parser` such as `tolerant` and `ecmaFeatures.jsx` (https://github.com/eslint/typescript-eslint-parser/blob/master/parser.js#L76-L83) and maybe some others?

josephfrazier:
Done, let me know what you think.

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,71 @@
...
+    'range',
+    'loc',
+    'tokens',
+    'comment',
+    'tolerant',
+    'useJSXTextNode',
+    {
+      key: 'ecmaFeatures',
+      title: 'ecmaFeatures',
+      fields: Object.keys(defaultOptions.ecmaFeatures),
```
RReverser: Can you show how this looks in UI please? I can't remember, just know that in acorn settings I had to flatten properties myself.

josephfrazier: Sure, it looks like this:

screen shot 2017-06-20 at 10 13 59 am

RReverser: Nice, I should use it in the other place as well then :) Didn't know nested options work (or maybe they didn't when I added it).

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,71 @@
...
+      settings:
+        settings => settings.ecmaFeatures || {...defaultOptions.ecmaFeatures},
+    },
+  ],
+};
```
RReverser: I think you need to pass `required=` with `range` option (see https://github.com/fkling/astexplorer/blob/a4edbaa49c1983ae4e0c9de2f460784f11afb099/website/src/parsers/lua/luaparse.js#L24), otherwise it will be possible to disable `range` option which will render highlighter useless (`nodeToRange` will not work at all).

Alternatively, check if `typescript-eslint-parser` returns `start` & `end` properties on the nodes (like Acorn-derived parsers do), in which case use them for `nodeToRange` instead of `.range` itself.

website/src/parsers/js/typescript-eslint-parser.js
josephfrazier: One caveat I noticed is that if you have a non-nested option after the nested one(s), it will also appear to be nested. Probably just a simple CSS fix needed, though.

RReverser: @josephfrazier I guess not as much CSS, as reordering of properties (render non-nested always first). Although I don't think it's a big deal, people can order properties theirselves when adding parsers, so should be fine.

website/src/parsers/js/typescript-eslint-parser.js
josephfrazier: Ah, good point. Fixed in https://github.com/fkling/astexplorer/pull/242/commits/4912cfb93ffbc52a9aea7d21e336522d9ee82bdf

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,72 @@
...
+    require(['typescript-eslint-parser'], callback);
+  },
+
+  parse(parser, code, options) {
+    return parser.parse(code, {...defaultOptions, ...options} );
+  },
+
+  nodeToRange(node) {
+    if (node.range) {
+      return [node.range[0], node.range[1]];
```
RReverser: Btw, can't just do `return node.range` here instead of entire condition?

josephfrazier: Heh, that's what I get for doing super-focused edits :P Fixed in https://github.com/fkling/astexplorer/pull/242/commits/49f9ff92dd1e9021cf117608f2bf74f916e21cae

fkling:
@RReverser: please merge whenever you think its ready (and add a the label "needs site push" so that I know that I have to push a new version 😉 )

RReverser:
@fkling I want to hold onto this just a little bit, give me time for final reviews :)

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,70 @@
...
+import SettingsRenderer from '../utils/SettingsRenderer';
+
+const ID = 'typescript-eslint-parser';
+
+const defaultOptions = {
+  range: true,
+  loc: false,
+  tokens: false,
+  comment: false,
+  tolerant: true,
```
RReverser: Perhaps should be `false` by default? (so that user would see proper syntax errors)

josephfrazier: Oh, I had copied that from [espree](https://github.com/fkling/astexplorer/blob/93aa4be41e00e0754afc97c0f9eebd26d401eaca/website/src/parsers/js/espree.js#L13). Should we change it there too? It looks like [esprima](https://github.com/fkling/astexplorer/blob/93aa4be41e00e0754afc97c0f9eebd26d401eaca/website/src/parsers/js/esprima.js#L14) and [recast] already (https://github.com/fkling/astexplorer/blob/93aa4be41e00e0754afc97c0f9eebd26d401eaca/website/src/parsers/js/recast.js#L11) default to `false`.

RReverser: Hmm, yeah, I would say it's a good idea to have visible errors by default.

RReverser:
@josephfrazier Travis production build failed - can you please fix it (by making sure dependency is transpiled, see `webpack.config.js`)?

website/src/parsers/js/typescript-eslint-parser.js
```diff
@@ -0,0 +1,70 @@
...
+
+  loadParser(callback) {
+    require(['typescript-eslint-parser'], callback);
+  },
+
+  parse(parser, code, options) {
+    return parser.parse(code, {...defaultOptions, ...options} );
+  },
+
+  nodeToRange(node) {
```
azz: The default implementation is the same (you can delete this)

josephfrazier:
@RReverser, this should be good to go now. Thanks!

RReverser:
@josephfrazier Yup, CI looks green. Thanks for your patience!

josephfrazier:
Thanks for the merge!

fkling:
Sorry for the long wait! I was busy with lots of other things ¯\_(ツ)_/¯

josephfrazier:
No problem, thanks for the push! Would it be possible to automate the release process? I'm not exactly sure how it's currently handled.

fkling:
> Would it be possible to automate the release process?

Only if we 

- Get long term caching to work properly *
- Forsake long term caching for the traditional "last-modified" approach

* I followed different tutorials for this, but basically the way webpack is setup now results in tiny changes to files (module IDs are getting updated) without updating the chunk hash, which means the content changes but not the file name. This breaks the app for people. In these cases I have to bump the cache breaker.
Maybe we could even automate this, if someone wants to invest the time 😉 

I'm thinking this might be less of a problem if we'd serve parser bundles completely separately from the website/UI itself ( #197).

> I'm not exactly sure how it's currently handled.

Depending on what changed I may or may not run `yarn build` and `yarn start` and look over astexplorer locally.

Then I run [`./scripts/build-site.sh`](https://github.com/fkling/astexplorer/blob/63dacad8b9113aea2629a917a71afdbe3a4ab3d3/scripts/build-site.sh) to create a production build in a separate branch (`gh-pages` for historical reasons). I switch to the branch, and check the commit whether there have been changes to files without changing their names.
If yes I revert the commit, switch back to master, bump the cache breaker and run the script again.

Then I run [`./scripts/deploy.sh server gh-pages`](https://github.com/fkling/astexplorer/blob/63dacad8b9113aea2629a917a71afdbe3a4ab3d3/scripts/deploy.sh) which pushes the branch to the server.

On the server I have a post-receive hook that looks like

```bash
#!/bin/bash

set -e

repo=/data/astexplorer/repo
website=/data/astexplorer/website
server=/data/astexplorer/server
foldername=$(date +%Y-%m-%d_%H-%M-%S)

while read line; do
  case "$line" in
    *refs/heads/{{website_branch}}*)
      git clone --branch {{website_branch}} --depth 1 "file://$repo" $website/$foldername
      ln -f -n -s $foldername $website/latest
    ;;

    *refs/heads/{{server_branch}}*)
      git clone --branch {{server_branch}} --depth 1 "file://$repo" $server/$foldername
      cd $server/$foldername/server/
      # GIT_DIR is set which breaks package installation 
      # env -i creates a new empty environment
      env -i yarn
      ln -f -n -s $foldername $server/current
      echo "Restarting astexplorer server..."
      sudo systemctl restart astexplorer
    ;;
  esac
done
```

which checks out the latest version the website into a new repository (or updates and restarts the server).

Then I go to `latest.astexplorer.net` checking again whether everything is fine. If yes, I ssh into the server and update a symlink to point to the latest version.

Any suggestions for making this easier are appreciated 😄

josephfrazier:
I'm not yet familiar enough with the webpack configuration to chime in on the caching issue, but it sounds like we could greatly simplify the process of reviewing and releasing pull requests:

With a few [changes](https://github.com/fkling/astexplorer/compare/master...e0b229a) (one of which is https://github.com/fkling/astexplorer/pull/265), I was able to make it so that astexplorer can be deployed on Heroku ([example](https://warm-bastion-33330.herokuapp.com/)). Heroku offers [Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps), which makes it possible to deploy a pull request into an isolated environment ([example](https://github.com/josephfrazier/astexplorer/pull/1)). Similarly, deploying a new version after merging can be configured to happen automatically: https://devcenter.heroku.com/articles/github-integration.

How would you feel about switching to Heroku for hosting? It looks like the current server is set up to use CloudFlare (https://github.com/fkling/astexplorer/issues/69), but I think that can be done with Heroku as well: https://support.cloudflare.com/hc/en-us/articles/205893698-Configure-Cloudflare-and-Heroku-over-HTTPS

fkling:
Thank your for investigating this @josephfrazier! I made some progress with a simplified version of #197. So if this works all out, PRs for updating parsers won't be needed anymore, and most of the problems should go away 😃 

But I will keep the heroku solution in mind!

PR #243, opened 2017-06-18
Fix ESLint warnings/errors

josephfrazier:
This eliminates all the ESLint warnings/errors from `./website`. It also adds a `.travis.yml` file to lint/build `./website`. You can see an example Travis build here: https://travis-ci.org/josephfrazier/astexplorer/builds/244288099

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: node_js
+node_js:
+  - node
+cache:
+  yarn: true
+  directories:
+    - website/node_modules
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+script:
+  - cd website
+  - yarn
+  - yarn lint
+  - yarn watch -- --no-watch
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ cache:
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH=$HOME/.yarn/bin:$PATH
-script:
+install:
   - cd website
   - yarn
+script:
   - yarn lint
   - yarn watch -- --no-watch
--- a/website/src/storage/index.js
+++ b/website/src/storage/index.js
@@ -1,5 +1,3 @@
-// global.__cache = cache;
-
 export default class StorageHandler {
   constructor(backends) {
     this._backends = backends;
--- a/website/src/storage/index.js
+++ b/website/src/storage/index.js
@@ -1,13 +1,4 @@
-const cache = {};
 // global.__cache = cache;
-
-function getFromCache(cacheKey) {
-  return cache[cacheKey];
-}
-
-function setInCache(cacheKey, revision) {
-  cache[cacheKey] = revision;
-}
 
 export default class StorageHandler {
   constructor(backends) {
--- a/website/package.json
+++ b/website/package.json
@@ -88,6 +88,7 @@
     "postcss-safe-parser": "^3.0.0",
     "postcss-scss": "^1.0.0",
     "prettier": "^1.3.1",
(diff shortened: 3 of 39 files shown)


RReverser:
Ha, also fixed there in my fork (https://github.com/RReverser/esprima_ast_explorer/commit/1fc97b939389e96e057a7d45e63b0fdbfddd77d1, https://github.com/RReverser/esprima_ast_explorer/commit/70bc51ccea84114bd784eb51cc10db610b4e1bfe, etc.). Guess I should start sending out PRs to avoid duplications, but this should be fine :)

website/src/storage/index.js
```diff
@@ -1,14 +1,5 @@
-const cache = {};
 // global.__cache = cache;
```
RReverser: Perhaps remove comment too?

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  directories:
+    - website/node_modules
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+script:
+  - cd website
+  - yarn
+  - yarn lint
+  - yarn watch -- --no-watch
```
RReverser: Why not `yarn build`? Just for performance reasons? I think it could be still more useful to make sure that production version is built correctly.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+cache:
+  yarn: true
+  directories:
+    - website/node_modules
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+script:
+  - cd website
+  - yarn
```
RReverser: I don't know if it makes any difference, but I think it would make sense to separate dependency installation into `install` section rather than part of `script`.

.travis.yml
josephfrazier: Yeah, the production build is considerably slower, so I used this instead to make PR feedback faster (assuming Travis is enabled for this repo). I can change it, though, if you'd still prefer `yarn build`. Just let me know.

RReverser: I guess it's fine for now, I'm working on unit tests, will change it myself if required.

RReverser:
Looks great, thanks!

RReverser:
Grouped/squashed into two commits and merged onto master.

RReverser:
@fkling I don't have rights to enable Travis on the repo, could you please?

josephfrazier:
Thanks!

fkling:
Done! https://travis-ci.org/fkling/astexplorer

Thank you all, this is great!

RReverser:
@fkling Can you also trigger build manually? (from Settings -> Integrations -> Travis on Github repo as far as I remember)

fkling:
@RReverser: The only thing I see is a "Test service" button. Is that what you mean?

RReverser:
@fkling Yup!

RReverser:
(it's just a one-time thing required in situations when Travis was enabled on repo with already-existing `.travis.yml`, to check that it works)

RReverser:
image

Heh, WebPack progress bar plugin doesn't seem to check whether output is TTY, but that's not a big deal.

RReverser:
Yay, it's green. https://travis-ci.org/fkling/astexplorer

josephfrazier:
Thanks everyone! It looks like the build is using a VM rather than a container, which is making it start slower than it could. I'll follow up with another PR.

screen shot 2017-06-20 at 11 20 25 am

RReverser:
@josephfrazier Perfect, thanks!

.travis.yml
fkling: > Yeah, the production build is considerably slower

One more reason to separate parsers/tools and the website...

PR #245, opened 2017-06-21
Update glimmer packages to 0.25.1.

rwjblue:
The format of the AST transforms changed, this adjusts to the new format.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -38,8 +38,8 @@
     "webpack-chunk-hash": "^0.4.0"
   },
   "dependencies": {
-    "@glimmer/compiler": "^0.22.0",
-    "@glimmer/syntax": "^0.22.0",
+    "@glimmer/compiler": "^0.25.1",
+    "@glimmer/syntax": "^0.25.1",
     "acorn": "^5.0.3",
     "acorn-jsx": "^4.0.1",
     "acorn-to-esprima": "^1.0.2",
--- a/website/src/parsers/handlebars/glimmer.js
+++ b/website/src/parsers/handlebars/glimmer.js
@@ -9,7 +9,7 @@ export default {
   id: ID,
   displayName: ID,
   version: pkg.version,
-  homepage: pkg.homepage || 'https://github.com/tildeio/glimmer',
+  homepage: pkg.homepage || 'https://github.com/glimmerjs/glimmer-vm',
 
   loadParser(callback) {
     require(['@glimmer/syntax'], (glimmer) => callback(glimmer.preprocess));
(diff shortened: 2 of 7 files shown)


RReverser:
Hmm I'm not sure what we want to do in these cases, given that there might be saved snippets that rely on the old API.

@fkling Do we really want to preserve old snippets working forever? We might easily blow up tools x versions matrix that way (no offence to the on-demand tool builder approach).

Turbo87:
@RReverser would it be possible to have both versions in parallel for a while and display some sort of warning for the old version?

rwjblue:
I believe that I could do `packages/glimmer-0-24` and `packages/glimmer-0-25`, but I'm not sure the gists have enough meta data to know which version they were built against...

RReverser:
Yeah, the old Glimmer currently would have to be preserved as the only "glimmer" to preserve compatibility and new one would have to be versioned...

fkling:
> Do we really want to preserve old snippets working forever?

Isn't that how the web is supposed to work? 😉 

Idk, I saw astexplorer similar to jsfiddle and you wouldn't want old jsfiddles to break would you?

Building older versions on demand would make this a bit better but of course it also introduces additional complexity.

Alternatively, since we are already storing the version of the parser, we could show a warning that the snippet was created with an older version and explain that results might differ or transforms might be broken. But overall that seems like a bad user experience.

RReverser:
@fkling 

> Idk, I saw astexplorer similar to jsfiddle and you wouldn't want old jsfiddles to break would you?

I guess here's the difference - I mostly see it as a debugging sort of tool that you might want to share with other people to reproduce some issues or play with transformer idea, but when it's done, never come back :) JSFiddle, on the other hand, is more for long-term visual demos rather than debugging sessions.

> So alternatively we could show a warning that the snippet was created with an older version and explain that results might differ or transforms might be broken. But overall that seems like a bad user experience.

Or we could auto-run codemods when such snippet is opened :)

fkling:
@RReverser 

> but when it's done, never come back

idk, I can imagine people referencing codemods in astexplorer in blog posts and such.

Anyways, I'm not saying that multi-version support is something we absolutely must implement now, but we shouldn't do anything that prevents us from adding it in the future.

fkling:
How should we proceed?

rwjblue:
@fkling - I need to update for @glimmer/compiler@0.26.x (which released a short while after I submitted this).

Also, I am happy to update to support versioned glimmer versions.

fkling:
If you don't want to support older versions that's OK for me. So, should I merge this before you make any other updates?

fkling:
@rwjblue: Could you update this PR to be compatible with the latest version of glimmer and then I'l;l merge it? Thank you!

rwjblue:
Absolutely! I’ll try to have this updated this week...

fkling:
Superseded by #287

PR #246, opened 2017-06-21
Add `prettier` transformer

josephfrazier:
This makes it possible to use astexplorer like https://prettier.github.io/prettier,
but with the ability to cross-reference the code with its AST.


website/webpack.config.js
```diff
@@ -26,15 +26,6 @@ const plugins = [
   // so that code patch will never be executed.
   new webpack.IgnorePlugin(/^eslint$/, /babel-eslint/),
 
-  // Prettier //
-
-  // typescript-eslint-parser is a dev dependency of prettier, so it's not
-  // installed by default but prettier does still require it. See
-  // https://github.com/prettier/prettier/issues/986
-  new webpack.IgnorePlugin(/typescript-eslint-parser/, /\/prettier/),
```
RReverser: If you want to upgrade Prettier, you need to provide new regexes to exclude irrelevant parsers (they're structured differently in Prettier now). Unless you want to allow all of the parsers, but then need to check that transform actually works with each of them.

azz:
> Unless you want to allow all of the parsers, but then need to check that transform actually works with each of them.

That would be great. With the next version of prettier there's an opportunity to [modify the AST](https://github.com/prettier/prettier#custom-parser-api) so AST Explorer would be the perfect place to try that out.

```js
// codeExample.txt
export default ({ traverse }) => ({
  parser(code, { babylon }) {
    const ast = babylon(code);
    traverse(ast, {
      Identifier(path) {
        path.node.name = path.node.name.split('').reverse().join('');
      }
    });
    return ast;
  }
});
```

website/webpack.config.js
josephfrazier: Ah, fixed in https://github.com/fkling/astexplorer/pull/246/commits/69276968c80cdb4f1a7c9c2eb7562043a427b3ab

Since this is a JavaScript/etc transformer, I eliminated the non-JS parsers.

RReverser:
@fkling I wonder if we can / should limit certain transformers to a list of supported parsers to make sure they're not invoked with something that produces an incompatible AST? (and how that would look like in the UI)

Maybe let parsers and transformers define the standard which they produce aka "estree", "babel", "typescript", etc. to filter them out by such group?

fkling:
@RReverser 

> Maybe let parsers and transformers define the standard which they produce aka "estree", "babel", "typescript", etc.

Regardless of anything else, I think that would be a  good idea.

> I wonder if we can / should limit certain transformers to a list of supported parsers

IMO the current structure (select parser or select transform + parser) is wrong and confusing. What I started in #197 was to unify parsers and transforms:

> **Unified interface**
>
> One big difference is that transformers also have to provide a `parse()` method (instead of specifying the ID of a parser to use). This method is supposed to delegate to the parser the transform is using. That means, whether a tool is a transformer only depends on whether or not it implements a transform method.
> Overall this should simplify the store logic and other places on the website.
Another positive side effect of this: The "parser settings" logic can now be reused for transformer settings.

The reasoning behind this is:

- At the moment, parsers and transformers are completely independent. I.e. any settings you make to a parser doesn't affect the transform at all. This is confusing. The version the transform is using internally might also be different than the version we provide.
- By delegating to the transformers own parser we can show an AST that is closer to what the transformer is actually using.
- Not all transformers allow you to change parser settings anyway.

Of course not all transformers expose the parser. In that case we have to bundle the right version and "replicate" the parsing process (as long as it's reasonable to do so).

I.e. in the end users wouldn't select a parser or a parser + transform anymore, but just a "tool". A tool exposes a `parse()` or a `parse()` and a `transform()` method.

fkling:
Is this good to go? Code looks fine to me.

josephfrazier:
Yup, I just resolved the merge conflict, and don't have any more changes to make.

PR #247, opened 2017-06-22
Add support for Scala

gabro:
This PR adds support for Scala, using the parser from scalameta, recently compiled to JS (thanks to ScalaJS) and published on npm!

The final JS api of the parser is still a work in progress (see https://github.com/scalameta/scalameta/pull/957) but it's close to be merged into scalameta.

As of now, I'm using a beta version I published on npm, that I think will drop the beta suffix as soon as scalameta 1.9.0 is released.

Here's a GIF of astexplorer in action with Scala 🎉


gabro:
Woops, I forgot to commit a bunch of font-related files that made the CI fail. Done now.

website/src/parsers/scala/index.js
```diff
@@ -0,0 +1,6 @@
+import 'codemirror/mode/clike/clike';
+
+export const id = 'text/x-scala';
```
RReverser: `id`s are usually meant to be same as folder name, this looks like was supposed to go to `mimeTypes` instead?

gabro: I had to use this in order to get the correct syntax highlight in codemirror. Is there any other way to set-up codemirror other than the category id?

RReverser: Oh, right. This is weird naming on CodeMirror side compared to other modes...

gabro: Yeah, it took me a while to figure it out and it's quite ugly as an id.
That said, it seems to be working :)

website/src/parsers/scala/scalameta.js
```diff
@@ -0,0 +1,81 @@
...
+  ],
+  required: new Set('dialect')
+}
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: ID,
+  version: '1.9.0',
```
gabro: Hu, i should probably get this from the npm package, I'll fix it soon

gabro:
I've updated the Scalameta dep, and the API is now stable.

@RReverser any other comments on this? :)

RReverser:
@gabro Just one - can you please make sure that syntax error correctly highlights the line? I've noticed some of the parsers in the repo don't and want to add a test against that, but for now a manual check would do.

If it's not, you'll need to do something like this:

```js
try {
  return smth.parse(...);
} catch (e) {
  e.lineNumber = ...line number from the error (might be in some other property already or will need to convert to a number from the message)...;
  throw e;
}
```

RReverser:
Or you can leave it as-is and I'll fix it myself later, but this would help a bit :)

RReverser:
Looks like scalameta-parsers returns `pos.{start, end}` on error objects which is somewhat hacky to convert to a line number (although we do that in few other parsers).

gabro:
I've been wondering about how to report errors, and I didn't know throwing an error with the `lineNumber` property was the way to go.

I can include the line number on the scalameta side, so that'd be easier.

RReverser:
@gabro Thanks, that would be great! Currently we check for several properties to cover most usual cases - `.loc.line`, `.lineNumber` and `.line` - so either should work.

gabro:
Waiting for: https://github.com/scalameta/scalameta/pull/962

RReverser:
LGTM, thanks! `columnNumber` is not currently used, but maybe will be useful in the future or for other scalameta consumers? Totally up to you how to expose it :)

gabro:
Cool! I've just upgrade the dep, and I think this is ready to go!

gabro:
Woooo 🙌 do you know when this will be live?

RReverser:
That's up to @fkling :) Maybe we'll wait for final fixes in https://github.com/fkling/astexplorer/pull/242 to include it too.

fkling:
Building and punishing is a manual process atm. I won't have time to it before tomorrow, sorry :-/

gabro:
No rush, don't worry. Thanks for the prompt response :)

--
gabriele
~ buildo

fkling:
Also lol autocomplete... of course meant "pushing" not "punishing". 😛

gabro:
Lol, I read it fast and I thought you said "building is a punishing process" 😅

fkling:
That is true too 😆

fkling:
Pushed!

fkling:
@gabro: Your [article](https://blog.buildo.io/exploring-scala-ast-in-your-browser-dc0b1fb743e0) is generating traffic 😃  Thanks for the shoutout! (cc @RReverser)
screen shot 2017-06-26 at 10 25 42 am

gabro:
Haha yes, it ended up in HN front page, so that may be the cause 😁

PR #249, opened 2017-06-25
Add typescript-eslint-parser

azz:
Hi guys, thanks for this awesome tool! It is incredibly useful when working on prettier.

`typescript-eslint-parser` is the parser prettier uses for TypeScript support, and it would be great if ASTExplorer could host it too.

The parser aims to be a superset of ESTree, so it should be pretty low-maintenance.


azz:
D'oh! didn't see #242. Closing.

Issue #248, opened 2017-06-22
Support generic languages in transform


gabro:
I'm playing around with a custom transformation in Scala (based on #247) and it seems like the source code of the transformation is required to be in JavaScript.

Is this the case, or I'm getting it wrong? If so, can this be generalized?

RReverser:
Well the transforms are executed on the client-side (in the browser), so yeah, they have to be in JavaScript. Or do you want to put there some language that transpiles to JavaScript? That should work, too, but probably will require some tweaking around highlighting.

gabro:
My idea would be to add scalafix, which can take source files defining transformations as an input.

Scalafix can be compiled to js, and be invoked in the browser. The only thing it would need is to access the content of the transform pane in its raw form.

RReverser:
> The only thing it would need is to access the content of the transform pane in its raw form.

That's definitely possible, you don't have to run `transformJSCode` on the transform code, it's just for convenience for transformers that really use JS.

fkling:
What @RReverser said. If you look at https://github.com/fkling/astexplorer/blob/93aa4be41e00e0754afc97c0f9eebd26d401eaca/website/src/parsers/js/transformers/jscodeshift/index.js#L47, `transformCode` is just a string. The transform can do with that whatever it wants.

gabro:
I've tried to implement a test transform where I simply return the original code, but I still get syntax errors if transformCode is not valid JS code. I'm literally just doing 'return code', is there any setting I've missed?

gabro:
At a first glance it seems like the TransformOutput component transforms the transformCode to ES5 before passing it to the transformer.
But I'm from the phone atm, so I might be getting it wrong

fkling:
Oh you are right... maybe this is a good time to change how transformers work on the website. Instead of the site providing the `TransformOutput` component, I was thinking that every transformer should be able to specify what "view" to render.

E.g. something like the prettier transformer which uses the transform input as configuration source could provide a more specialized JSON editor. Or eslint could render two panels, one for errors and one for displaying the autofixed source.

Transformers who accept JavaScript as the transform could then use a special component that does the compilation.

gabro:
That would make a lot of sense. A smaller step towards a less JS-only solution would be to also pass the raw input to the transformer. That's a small but very significant change, would you accept a PR for it?

fkling:
>  A smaller step towards a less JS-only solution would be to also pass the raw input to the transformer.

[...]

PR #250, opened 2017-06-26
Avoid compiling to ES5 in TransformOutput component

gabro:
As discussed in #248, `TransformOutput` automatically tries to compile the transform code to ES5.

This makes it impossible to provide transformers in languages other than JavaScript.

This PR extracts the `toES5` step into a separate utility, to be explicitly invoked by transformers that need it. I've provided an example of what it would look like for the `babel` transformer. If the approach is deemed ok, I'll port this to all the existing transformers.

### DIFF
--- a/website/src/components/TransformOutput.js
+++ b/website/src/components/TransformOutput.js
@@ -3,44 +3,20 @@ import Editor from './Editor';
 import JSONEditor from './JSONEditor';
 import PropTypes from 'prop-types';
 import React from 'react';
-import halts, {loopProtect} from 'halting-problem';
 import {SourceMapConsumer} from 'source-map/lib/source-map-consumer';
 
 import stringify from 'json-stringify-safe';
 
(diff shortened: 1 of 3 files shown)


RReverser:
Should this maybe merged into `compileModule` or there are cases when we need just one of them? Other than that, I liked this idea.

gabro:
Ah yes, that's probably a good idea. I'll get to it later today.

gabro:
The only significant change is that `compileModule` would return a Promise now.

fkling:
How should we proceed?

gabro:
Hey @fkling, sorry for leaving this hanging. If there's consensus I can wrap this up and move the logic to `compileModule`, making it return a promise.

How does it sound?

fkling:
FWIW, this was superseded by #290. It should be possible to use transformers in different languages now.

PR #253, opened 2017-07-12
Add PHP parser by ichiriac

pornel:


website/src/parsers/php/php-parser.js
```diff
@@ -0,0 +1,40 @@
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'php-parser/package.json';
+
+const ID = 'php-parser';
+
+const defaultOptions = {
+  parser: {
+    extractDoc: true,
```
RReverser: Should this be configurable in the UI?

website/src/parsers/php/php-parser.js
```diff
@@ -0,0 +1,40 @@
...
+  loadParser(callback) {
+    require(['php-parser'], callback);
+  },
+
+  parse(Engine, code, options) {
+    const parser = new Engine({...defaultOptions, ...options});
+    return parser.parseCode(code, '');
+  },
+
+  getNodeName(node) {
```
RReverser: Please implement `nodeToRange` for node highlighting and "jumps" to work.

website/src/parsers/php/php-parser.js
pornel: This adds doc comments as ast nodes, so it doesn't make a big difference.

website/src/parsers/php/php-parser.js
```diff
@@ -0,0 +1,40 @@
...
+  homepage: pkg.homepage,
+  locationProps: new Set(['loc']),
+
+  loadParser(callback) {
+    require(['php-parser'], callback);
+  },
+
+  parse(Engine, code, options) {
+    const parser = new Engine({...defaultOptions, ...options});
+    return parser.parseCode(code, '');
```
RReverser: Does error line highlighting work? If it doesn't currently, please see https://github.com/fkling/astexplorer/pull/247#issuecomment-310726688 for some tips on how to add that.

pornel: Error highlighting works.

website/src/parsers/php/php-parser.js
pornel: Done

website/src/parsers/php/php-parser.js
RReverser: Great, thanks for checking.

fkling:
Seems fine to me!

fkling:
It throws an error in the production build because uglify is renaming properties. Have to investigate how to fix that.

fkling:
Even if I turn mangling off it throws an error:
screen shot 2017-07-19 at 10 27 47 pm
It seems that `this` is defined as 

screen shot 2017-07-19 at 10 28 34 pm

Not sure what exactly the minifier does that makes the `variable` property not defined.

fkling:
Disabling minification for anything file matching `/php-parser/` works for now...

RReverser:
> because uglify is renaming properties

Wait what? It doesn't do that by default, and I don't remember us configuring it...

fkling:
@RReverser: See the following comments 😉 I assumed it was properties, because that's how it looked like first. But even with mangle turned off completely it's still broken.

PR #263, opened 2017-08-26
Fix typescript parser's node name

koczkatamas:
Reverse SyntaxKind lookup does not give correct results (as described here: https://github.com/Microsoft/TypeScript/issues/18062) . This commit workarounds that issue (until it'll be resolved in TS compiler - if it will be ever...)


website/src/parsers/js/typescript.js
```diff
@@ -64,6 +65,14 @@ export default {
     }, compilerHost);
 
     const sourceFile = program.getSourceFile(filename);
+    
+    // workarounds issue described at https://github.com/Microsoft/TypeScript/issues/18062
+    for (const name of Object.keys(ts.SyntaxKind).filter(x => isNaN(parseInt(x)))) {
```
RReverser: This should happen in `loadParser` and not on each `parse` invocation.

koczkatamas: Fixed. Thanks!

PR #265, opened 2017-08-28
Use uglify-es instead of submodule

josephfrazier:
uglify-es is now released from the "harmony" branch of
https://github.com/mishoo/UglifyJS2.git, so it should be equivalent.

See https://github.com/mishoo/UglifyJS2/tree/067d52b6bac67378cccbc8fdea2320676aafc47e


fkling:
Thank you! Love it when things get simpler 😃

Issue #267, opened 2017-08-29
Add Babylon 7 TypeScript option


thejameskyle:
There's now a "typescript" plugin to Babylon in 7.0 https://github.com/babel/babylon#plugins

PR #268, opened 2017-08-29
Add TypeScript option to Babylon 7

thejameskyle:
Resolves https://github.com/fkling/astexplorer/issues/267

### DIFF
--- a/website/src/parsers/js/babylon7.js
+++ b/website/src/parsers/js/babylon7.js
@@ -19,6 +19,7 @@ const availablePlugins = [
   'dynamicImport',
   'numericSeparator',
   'importMeta',
+  'typescript',
 ];
 
 const ID = 'babylon7';


xtuc:
Flow annotations will collide with TypeScript syntax, it might be annoying for the peoples used to Flow in astexplorer.

What about two different parser? One _Babylon7-Flow_ with the Flow flag for Babylon and _Babylon7-TypeScript_ which would use the TypeScript flag.

thejameskyle:
Only if you have both enabled which Babylon should throw an error which AST Explorer can handle as it does already. We should fix this on the Babylon side and leave AST Explorer (and any other consumer) with a simpler implementation.

hzoo:
^ Babylon should already throw an error when you try to turn on both so this should be fine

thejameskyle:
@fkling do you need anything for this?

hzoo:
I think he may be out/busy still (vacation)?

fkling:
I'll update everything tomorrow!

PR #275, opened 2017-09-30
Add Hide type keys option

xixixao:
The type keys seem redundant, added an option to hide them, making the AST more readable:
screenshot 2017-09-30 22 19 53


fkling:
Seems reasonable. Not every AST uses `type` in such a way (e.g. Typescript actually stores data type information there), but in such a keys the option can just be disabled.

PR #279, opened 2017-10-14
Add ability to customize key mappings to editor

NickTomlin:
First, thanks so much for this project! It's made my journey diving into the world of ASTs so much easier.

This pull request adds the ability to customize the key mapping for the `Editor` and `JSCodeShiftEditor` CodeMirror instances with mappings for `emacs`, `sublime`, and (last but not least) `vim`.

screen shot 2017-10-14 at 11 15 33 am

keymaps-demo-loop

Things to note:
---

- I went ahead and just stuck another button on the top row; this works but I'd be open to exploring a more meta "options page" if that's a good fit
- Because `Editor` and `JSCodeShiftEditor` share the same class but have different containers I had to duplicate some prop passing; let me know if there's a better way
- I'm not a fantastic designer, let me know if there's a preferred pattern for this UI

Things I tested:
---
- [x] Switching between parser types (css/js/php)
- [x] Switching between transformers
- [x] Switching between bindings (default -> vim -> default -> emacs)
- [x] Reloading (setting is persisted)

Size
---

Also, including these additional mappings does increase the build size for app but it is not significant IMHO:

```
App.js Size
---

before: app-7c94119c6a8277cce7d2-12.js     255 kB                50  [emitted]  [big]  app
after: app-badfe3b7918172f4e390-12.js     258 kB                50  [emitted]  [big]  app

Total Bytes (`perl -le 'map { $sum += -s } @ARGV; print $sum' -- ../out/*.js`)
---

before: 29016829
after: 29121463
```

Thanks!

### DIFF
--- a/website/src/components/Editor.js
+++ b/website/src/components/Editor.js
@@ -1,4 +1,7 @@
 import CodeMirror from 'codemirror';
+import 'codemirror/keymap/vim';
+import 'codemirror/keymap/emacs';
+import 'codemirror/keymap/sublime';
 import PropTypes from 'prop-types';
 import PubSub from 'pubsub-js';
 import React from 'react';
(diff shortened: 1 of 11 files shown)


fkling:
Sorry for not getting back to your earlier! This is pretty cool, I like it and everything looks good 😃

NickTomlin:
Not a worry, thanks!

PR #280, opened 2017-10-16
Update webidl2

TimothyGu:


fkling:
We may have to update the build configuration to transpile webidl as well, but let me update to babel-env first and see if it's necessary.

PR #287, opened 2017-11-20
Update `@glimmer` packages to v0.30.2

Turbo87:
Addresses https://github.com/fkling/astexplorer/pull/245#issuecomment-344621867

Closes #245

/cc @rwjblue @fkling

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -38,8 +38,8 @@
     "webpack-chunk-hash": "^0.5.0"
   },
   "dependencies": {
-    "@glimmer/compiler": "^0.22.0",
-    "@glimmer/syntax": "^0.22.0",
+    "@glimmer/compiler": "^0.25.1",
+    "@glimmer/syntax": "^0.25.1",
     "acorn": "^5.0.3",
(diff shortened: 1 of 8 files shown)


Turbo87:
@fkling it looks like the `simple-html-tokenizer` subdependency of `@glimmer/syntax` is not passed through Babel and so it still has ES6 code in it that throws when minified using UglifyJS. 

Any idea how to resolve this issue? I'm not exactly an expert on webpack 🤔

fkling:
> Any idea how to resolve this issue?

Everything that should be passed through babel [is configured here](https://github.com/fkling/astexplorer/blob/816ac89bb02cb3e27e5b84e6019261a556282d09/website/webpack.config.js#L142-L163). It might be as simple as adding

```js
path.join(__dirname, 'node_modules', 'simple-html-tokenizer'),
```

Turbo87:
@fkling indeed, that was all that was needed, thanks! :)

fkling:
Merged locally.

Issue #288, opened 2017-11-24
add TSLint support


cowchimp:
I'm new to the Typescript ecosystem but [TSLint](https://palantir.github.io/tslint/) seems to be a pretty popular Typescript linter.
Would be great to be able to experiment with it as part of the "Transform" tab.

fkling:
Merged.

PR #289, opened 2017-11-26
Add TSLint support

cowchimp:
Hi.

This PR adds [TSLint](https://palantir.github.io/tslint/) (the Typescript Linter) as a transformer (issue #288).
Use it with the rule example that's in `tslint/codeExample.txt` to see it in action.

Right now though the rule itself can only be written in Javascript, not Typescript.
Ideally I'd like to support writing the rules in Typescript, as most TSLint users are accustomed to.
But I think this will require a change in how the `transformer.transform` is called.
Right now the original `transformCode` is turned into `es5Code` before the specific TSLint transformer is called, and by that point that conversion throws an error.

LMK if you need more info\changes.
Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ are included so you can prototype your own plugins:
   - [babel][] (v5, v6)
   - [ESLint][] (v1, v2, v3)
   - [jscodeshift][]
+  - [tslint][]
 - CSS
   - [postcss][]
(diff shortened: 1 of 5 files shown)


fkling:
Yeah, the whole transform code parsing needs some overhaul 😉

PR #290, opened 2017-12-01
Allow writing Transformers in non-JS languages

cowchimp:
Hi.

As mentioned in #289 this is my attempt at a refactor that allows *not* applying babel to the `transformCode` for a given transformer. This is needed in order to allow writing transform code in other languages other than Javascript (e.g. a TSLint rule in Typescript).
I added a `transpilers` directory that currently contains two transpilers (`babel` and `typescript`). More can be added in the future.

I wanted to go ahead and present a working version that already applies this change to all of the existing transformers to make it easy to discuss\merge, but by-all-means I'd be happy to get feedback and change accordingly.
For example, you might prefer that in all existing transformers I load `transpiler/babel` via a regular `import` statement and not asynchronously via the `require` call inside `loadTransformer`, etc.


website/src/parsers/js/transformers/tslint/index.js
```diff
@@ -12,13 +12,26 @@ export default {
   defaultParserID: 'typescript',
 
   loadTransformer(callback) {
-    require(['tslint/lib/index'], tslint => callback({tslint}));
+    require([
```
tbranyen: Were there any other transformers that could benefit from being written in TypeScript?

cowchimp:
Merged upstream repo into fork.
Change `babel-plugin-macros` which was added after the PR was created.

cowchimp:
build now passes after merging upstream repo into fork.

fkling:
I have to admit that I have mixed feelings about the changes the transformers. On one hand, I feel it makes the transformer code more "noisy", on the other hand, I like the flexibility that this provides to transformers.

I'll merge this since I think this is the best we can do at the moment. After we have restructured a bit how parsing and transforming works in general, we may find ways to improve this.

(Please don't get me wrong, I'm not implying that your code is bad, I'm just wondering if there is something we can do to hide the transpiler from the transformers).

Since I'm going on vacation for two weeks, I might not merge this into stable (i.e. https://astexplorer.net). However, an unknown fact is that there is https://latest.astexplorer.com, which I use to test changes first and it runs the exact same build (process) except on a different domain. So if you want to use/test it, you can do that there.

PR #300, opened 2018-02-07
Adds missing uglify-js peer dependency of uglify-loader

cspotcode:
Build was failing because uglify-loader could not find uglify-js.


fkling:
Mmh, I wonder why it works for me though... anyways, it should go in `devDependencies` then.

PR #301, opened 2018-02-12
Add babylon7 as an available parser for recast

elliottsj:
screen shot 2018-02-11 at 4 03 32 pm

This allows parsing TypeScript sources when the babylon@7 'typescript' plugin is used:
screen shot 2018-02-11 at 4 01 47 pm

Once a [new version of jscodeshift is released to support TypeScript](https://github.com/facebook/jscodeshift/issues/180#issuecomment-364800400), this will be useful as a tool to author TypeScript transforms.

### DIFF
--- a/website/packages/babylon7/package.json
+++ b/website/packages/babylon7/package.json
@@ -4,6 +4,6 @@
   "description": "",
   "main": "index.js",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


blaster151:
Very eager to be able to use this enhancement!

PR #313, opened 2018-05-04
Add cherow js parser

fkleuver:


fkling:
cherow has been added in #366

PR #315, opened 2018-05-12
Fix jscodeshift with flow parser

xadn:
Using JScodeshift with the Flow parser broke due to the jscodeshift upgrade to v0.5.0 on April 2nd (a8344e10). 

This fix commit fixes it by passing the selected parser through to the transformer function, but I could have also simply loaded it as part of the loadTransformer function (similar to the babelv7 transformer).

Unexpected token error after the upgrade:
screen shot 2018-05-11 at 5 05 36 pm

With the fix:
screen shot 2018-05-11 at 5 05 39 pm

### DIFF
--- a/website/src/parsers/js/transformers/jscodeshift/index.js
+++ b/website/src/parsers/js/transformers/jscodeshift/index.js
@@ -11,10 +11,10 @@ export default {
   version: pkg.version,
   homepage: pkg.homepage || 'https://github.com/facebook/jscodeshift',
 
-  defaultParserID: 'recast',
+  defaultParserID: 'flow',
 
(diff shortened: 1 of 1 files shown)


xadn:
Ah ok, I can make the change much more self contained by defaulting to the
flow parser. I’m not aware of any downsides to using the flow parser
instead of recast, we just get the benefit of being able to handle the Flow
syntax.

I bisected the recent changes and the flow syntax did work before the
commit mentioned above.

xadn:
Closing since `export const parser = 'flow';` works. I might look into adding it to the template somehow so it's more discoverable.

PR #318, opened 2018-05-17
rename babel-macros to babel-plugin-macros

kentcdodds:
I did this out of an effort to reduce confusion (because `babel-macros` is what it used to be called, but it was renamed a few months ago).

I also significantly improved the `codeExample.txt` with more information on what was possible as well as some other tips.

screen shot 2018-05-17 at 9 32 49 am

screen shot 2018-05-17 at 9 31 56 am


kentcdodds:
Thanks! In case you're unaware, the site is not rendering due to a react error I think.

kentcdodds:
Whoops, you're right. Sorry. And thanks!

PR #329, opened 2018-06-24
Add yaml-unist-parser

mohsen1:
/cc @ikatyang

### DIFF
--- a/website/src/parsers/yaml/yaml-unist-parser.js
+++ b/website/src/parsers/yaml/yaml-unist-parser.js
@@ -11,11 +11,11 @@ export default {
   version: pkg.version,
   homepage: pkg.homepage || 'https://www.npmjs.com/package/yaml-unist-parser',
 
-  _ignoredProperties: new Set(['parent', 'errors', 'type']),
+  _ignoredProperties: new Set(['type']),
   locationProps: new Set(['position']),
 
   nodeToRange(node) {
-    if (node.position)
+    if (node.position && node.type !== 'null')
       return [node.position.start.offset, node.position.end.offset]
   },
 
--- a/website/src/parsers/yaml/yaml-unist-parser.js
+++ b/website/src/parsers/yaml/yaml-unist-parser.js
@@ -1,0 +1,35 @@
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'yaml-unist-parser/package.json';
+
+const ID = 'yaml-unist-parser';
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: ID,
+  version: pkg.version,
+  homepage: pkg.homepage || 'https://www.npmjs.com/package/yaml-unist-parser',
+
+  _ignoredProperties: new Set(['parent', 'errors', 'type']),
+  locationProps: new Set(['position']),
+
+  nodeToRange(node) {
+    if (node.position)
+      return [node.position.start.offset, node.position.end.offset]
+  },
+
+  getNodeName(node) {
+    return node.type;
+  },
+
+  loadParser(callback) {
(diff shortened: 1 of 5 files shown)


ikatyang:
FYI, I just published a new version with a bugfix for [multiline plain end position](https://github.com/ikatyang/yaml-unist-parser/pull/30).

website/src/parsers/yaml/yaml-unist-parser.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: ID,
+  version: pkg.version,
+  homepage: pkg.homepage || 'https://www.npmjs.com/package/yaml-unist-parser',
+
+  _ignoredProperties: new Set(['parent', 'errors', 'type']),
```
ikatyang: There's no `errors` prop and the `parent` is defined as non-enumerable prop (internal usage) so I guess we don't need to specify it here?

website/src/parsers/yaml/yaml-unist-parser.js
```diff
@@ -0,0 +1,35 @@
...
+  locationProps: new Set(['position']),
+
+  nodeToRange(node) {
+    if (node.position)
```
ikatyang: We probably need to exclude `Null` node here since its position is always `-1`.

mohsen1: is that when `node.type = "null"`?

ikatyang: https://github.com/ikatyang/yaml-unist-parser/blob/13fd6fb4ccbb465399a165864bde1312a2461755/src/types.ts#L89-L93

website/src/parsers/yaml/yaml-unist-parser.js
j-f1: Since there’s an option in the UI to hide the `type` key, should we always be hiding it?

website/webpack.config.js
```diff
@@ -255,6 +256,7 @@ module.exports = Object.assign({
       // I assume the `require(...)` call "succeeds" because 'fs' is shimmed to
       // be empty below.
       /flow-parser\/flow_parser\.js/,
+      /prettier/,
```
mohsen1: this breaks the runtime (`require` is not defined error) but not having it messes the uglifier.

fkling: Yeah, this has been an issue for a while. The runtime is already broken anyway. I have to look into this more. I'll merge this locally first later today.

website/src/parsers/yaml/yaml-unist-parser.js
```diff
@@ -0,0 +1,35 @@
...
+    if (node.position && node.type !== 'null')
+      return [node.position.start.offset, node.position.end.offset]
+  },
+
+  getNodeName(node) {
+    return node.type;
+  },
+
+  loadParser(callback) {
+    require(['yaml-unist-parser'], function (yamlAstParser) {
```
fkling: I think this can just be

    require(['yaml-unist-parser'], callback);

mohsen1:
Thanks! Please update the package while merging.

fkling:
I wanted to merge this, but when the parser is run on the default YAML code example, I'm getting the error

> cstItem.valueRange is Null

in `sequenceItems`. cc @ikatyang.

ikatyang:
Updating the parser should fix the issue ([ref](https://github.com/mohsen1/astexplorer/pull/1/commits/724222e9cb1613cba1df2c1601461a5139874996)).

fkling:
This was with version 1.0.0, but I will double check.

PR #331, opened 2018-07-09
Add ecmaVersion option 9, 10 for acron

othree:
acorn now support ecmaVersion 9 and 10.
And it need at least 9 to support for await of syntax.

### DIFF
--- a/website/src/parsers/js/acorn.js
+++ b/website/src/parsers/js/acorn.js
@@ -60,7 +60,7 @@ export default {
   _getSettingsConfiguration() {
     return {
       fields: [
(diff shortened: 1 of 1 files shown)


fkling:
I think the version currently available on the site doesn't understand `10` yet, but I can update that easily.

As for `await`: `async/await` is part of ES2017, so 8, that should work fine already.

othree:
Yes, you're right.
Actually, I am looking for `for await of` syntax support, not async, await.

jdalton:
Thanks for this PR! The site doesn't appear to have gotten the fix yet.

PR #339, opened 2018-08-20
flow-parser v0.79.1

panagosg7:


panagosg7:
I'll re-submit without the merge commit.

PR #342, opened 2018-09-13
[babel-plugin-macros] Add "decoratorsBeforeExport" option when specifying decorators plugin

skevy:
The "babel-plugin-macros" transform is currently erroring with: 

`The 'decorators' plugin requires a 'decoratorsBeforeExport' option, whose value must be a boolean.`

This should fix that.


fkling:
Thank you! There is at least the pipeline operator that needs to be updated as well, but I'll do this myself separately.

fkling:
I won't deploy this to stable until I'm back from vacation. Please use https://latest.astexplorer.net/ in the meantime.

PR #344, opened 2018-09-19
Add MDX

### LINKED ISSUE #343: New language: MDX
I would like to try to add support for [MDX][]. On one side you would write MDX content, on the other side you would get a MDX HAST tree. This would be super useful for debugging MDX, both for contributors and users.

Would you accept this? If yes, do you know how I should categorize this? Should I add it as a separate language next to Markdown? Or as a Markdown parser next to remark? Or as a remark transform? Maybe a separate Markdown parser makes most sense because MDX is heavily based on Markdown (it's even initially parsed as such), but it's not a remark transform, it just uses remark under the hood, but later it might use something else.

Thanks!

[MDX]: https://mdxjs.com/

silvenon:
- adds new file format MDX
- adds two parsers for MDX: mdxhast
- adds a transformer for MDX, which shows JSX output

I tried to make it as useful as possible for debugging, but the fact is that mdxhast doesn't really exist as a parser, it's just a useful level of debugging for @mdx-js/mdx. Is that ok? I made it just link to @mdx-js/mdx documentation.

Closes #343.

### DIFF
--- a/website/src/parsers/mdx/mdxhast.js
+++ b/website/src/parsers/mdx/mdxhast.js
@@ -1,0 +1,56 @@
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from '@mdx-js/mdx/package.json';
+import mdxAstToMdxHast from '@mdx-js/mdx/mdx-ast-to-mdx-hast'
+
+const ID = 'mdxhast';
+
+function removeNewlines(node) {
+  if (node.children != null) {
+    node.children = node.children.filter(node => node.value !== '\n');
+    node.children.forEach(removeNewlines);
+  }
+}
+
+export default {
+  ...defaultParserInterface,
+
+  id: ID,
+  displayName: ID,
+  version: pkg.version,
+  homepage: 'https://mdxjs.com',
+  locationProps: new Set(['position']),
+
+  loadParser(callback) {
+    require(['@mdx-js/mdx'], callback);
+  },
+
+  parse(mdx, code) {
+    let result;
+
+    mdx.sync(code, {
+      hastPlugins: [
+        mdxAstToMdxHast,
+        () => removeNewlines,
+        () => tree => {
+          result = tree;
+        },
+      ],
+    });
+
+    return result;
+  },
+
+  nodeToRange({ position }) {
+    if (position) {
+      return [position.start.offset, position.end.offset];
+    }
+  },
+
+  getNodeName(node) {
+    return node.type;
+  },
+
+  opensByDefault(node, key) {
+    return key === 'children';
+  },
+};
--- a/website/src/parsers/mdx/transformers/mdx/index.js
+++ b/website/src/parsers/mdx/transformers/mdx/index.js
@@ -1,0 +1,36 @@
+import compileModule from '../../../utils/compileModule';
+import pkg from '@mdx-js/mdx/package.json';
+
+const ID = 'mdx';
+
+export default {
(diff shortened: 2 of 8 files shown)


website/src/parsers/mdx/transformers/mdx/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+const ID = 'mdx';
+
+export default {
+  id: ID,
+  displayName: ID,
+  version: pkg.version,
+  homepage: 'https://mdxjs.com',
+
+  defaultParserID: 'MDXHAST',
```
silvenon: I'm not sure that I understand what this is for, so I'm not sure if I'm using it correctly. Could someone clarify?

website/src/parsers/mdx/mdxast.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+
+  parse(mdx, code) {
+    let result;
+
+    mdx.sync(code, {
+      mdPlugins: [
+        mdAstToMdxAst,
+        () => tree => {
+          result = tree;
```
silvenon: Even though I make the `mdx.sync` call, I only use the MDXAST tree. This is where I'm setting it, following how mdx works under the hood.

website/src/parsers/mdx/mdxhast.js
```diff
@@ -0,0 +1,56 @@
...
+
+  parse(mdx, code) {
+    let result;
+
+    mdx.sync(code, {
+      hastPlugins: [
+        mdxAstToMdxHast,
+        () => removeNewlines,
+        () => tree => {
+          result = tree;
```
silvenon: Same for MDXHAST.

website/src/parsers/mdx/mdxhast.js
```diff
@@ -0,0 +1,56 @@
...
+    require(['@mdx-js/mdx'], callback);
+  },
+
+  parse(mdx, code) {
+    let result;
+
+    mdx.sync(code, {
+      hastPlugins: [
+        mdxAstToMdxHast,
+        () => removeNewlines,
```
silvenon: This is only here to reduce noise. Each blank line turns into a text node and it's really noisy to read, I don't think it's useful at all. So I recursively remove all text nodes containing `\n`.

silvenon:
> Sorry for letting this sit for so long.

No problem! You have your life 😉

> I'm happy to merge this if this is still good to go.

I just updated some things, resolved the conflict and threw in Prettier, because raw JSX output from MDX is hard to read. Feel free to merge!

timneutkens:
@fkling we'd love to have this available, it saves a ton of time vs manually debugging output 🙏

timneutkens:
Awesome @fkling 💯

silvenon:
Yeeeeees! Thank you @fkling!

fkling:
Mmh, I don't know why it doesn't show in the menu at https://latest.astexplorer.net/ . I remember this happening before but I can't remember what the reason was. I see it in my local build.

Will have to take a look tomorrow, sorry!

silvenon:
Maybe it has something to do with localStorage, try clearing it. I didn't have to do anything to see it, though 😕

fkling:
No idea why it didn't appear for me. I even tried it in a private instance.

Either way, I bumped the cache breaker and it works now 😃

timneutkens:
💯 So happy with this 😌

PR #348, opened 2018-09-30
Fix prettier integration in the editor

satya164:


website/src/components/Editor.js
```diff
@@ -95,12 +95,14 @@ export default class Editor extends React.Component {
 
     this._bindCMHandler('blur', instance => {
       if (!this.props.enableFormatting) return;
-      require(['prettier'], prettier => {
+
+      require(['prettier/standalone', 'prettier/parser-babylon'], (prettier, babylon) => {
```
satya164: I have no idea what syntax this is. Never saw require taking a callback xD

RReverser: You never used AMD / Require.js?

satya164: I have seen it in libs but browserify was already a thing when I started coding xD
Didn't think it's being used here since this used webpack.

fkling:
Thank you. It's pushed to https://latest.astexplorer.net/ for now.

fkling:
Now also pushed to https://latest.astexplorer.net/ !

PR #354, opened 2018-11-12
Allow auto-fixers with ESLint v4

### LINKED ISSUE #123: transform for doing auto-fixes with ESLint?
Would it be possible to add a second mode for Eslint that just runs the eslint fix on the source code.

Being able to write Eslint plugins in ASTExplorer is awesome. This would make it possible to write functions to fix some of the issues automatically.

jridgewell:
This makes it much nicer to test a new auto-fix plugin.
Fixes #123.


jridgewell:
Ping @fkling: Could I get a review?

fkling:
Looks fine to me, thank you! I was thinking that it would be nice to update the default example to include a fixer, but this could also be done in a separate diff.

PR #356, opened 2018-11-23
Add auto-fixing to eslint v4 example transformer

jridgewell:
Re: https://github.com/fkling/astexplorer/pull/354#issuecomment-440574641


fkling:
Great, thank you!

PR #377, opened 2019-01-22
feat: add @typescript-eslint/parser

ikatyang:
Ref: https://eslint.org/blog/2019/01/future-typescript-eslint#what-this-means-going-forward


fkling:
Great, thanks you! We can probably remove the old one, but we can also do this whenever.

vivekgalatage:
For using certain features like import statements, the `@typescript-eslint/parser` requires the options to be configured properly such as `{ sourceType: "module", ecmaVersion: 6 }`

What is the way in which these options can be passed on https://astexplorer.net/

PR #389, opened 2019-03-30
Remove unnecessary getNodeName overrides

RReverser:
All of these have exactly same implementations as the base one and only make it harder to find real custom overrides.

### DIFF
--- a/website/src/parsers/css/csstree.js
+++ b/website/src/parsers/css/csstree.js
@@ -22,10 +22,6 @@ export default {
     }));
   },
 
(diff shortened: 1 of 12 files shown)


fkling:
Thank you for all the improvements!

I'm working on some changes that decouple the actual parser wrapper from AST traversal stuff. I planned to do this as part of a larger rewrite, but maybe it's easier to introduce the changes little by little.

RReverser:
That sounds great and should help a lot.

PR #396, opened 2019-04-13
Add Rust parsing via syn

RReverser:
This took a while, but I wrapped a popular Rust parser [`syn`](https://github.com/dtolnay/syn) into a custom crate [`astexplorer-syn`](https://github.com/RReverser/astexplorer-syn) that compiles the actual parser to WASM and provides conversion of all AST nodes to JavaScript that tries to resemble Rust interface as closely as possible while exposing all the useful data.

Also I've converted an official SVG logo via fontcustom and added to the repo.

Long story short, you can now parse Rust code in the browser:
image

This is current parsing Rust code as a whole file. In the wrapper I've also exposed an API to parse derive inputs, and we can still expose them later via options, but I'm not sure how useful they are without transform capability which is more complicated to implement.


fkling:
Looks all good to me 😃 Can't comment on the icon/font stuff though 😆 The `this.lineOffsets` business will become a bit nicer with the changes I have in mind (but no idea when I'll have time to work on it).

fkling:
I had to add the `application/wasm` mime-type to the server config, but now it works 😃 Will post on Twitter tomorrow.

RReverser:
Great, thanks!

Btw, another thing I've been thinking about - would it be worth to enable some of the "Hide ..." options by default? Or at least remember them across sessions? Currently it's somewhat annoying to toggle them every time I open ASTExplorer.

fkling:
You are not the first one to point that out 😉 It should be easy to support this, even in a "hacky" way.

fkling:
FWIW, I added storing tree filter settings: 488a092fbd33ce60c1863c3de8878732767f55b2

PR #400, opened 2019-05-04
Add support for ReasonML & OCaml

### LINKED ISSUE #350: ReasonML/OCaml support
This is a feature request.

Requesting support for ReasonML support which uses OCaml as a supporting background programming language which means support for OCaml can come out of the box as well.

[BuckleScript](https://github.com/BuckleScript/bucklescript) is the translation layer currently binding OCaml compiler to javascript. So AST can be handled the same way they're handling it probably. (**Just a suggestion**)

I'd be also happy to pick it up and work on it if someone can help me with it.

jchavarri:
Fixes #350.

Adds support for OCaml and ReasonML through refmt.

I added a couple of minor changes to the more general code, will comment inline.

@fkling Please let me know if there's anything missing! 🙂


website/src/components/Toolbar.js
```diff
@@ -8,7 +8,7 @@ import KeyMapButton from './buttons/KeyMapButton';
 
 export default function Toolbar(props) {
   let {parser, transformer, showTransformer} = props;
-  let parserInfo = parser.id;
+  let parserInfo = parser.displayName;
```
jchavarri: In Reason & OCaml case, the parser is shared, but it can't have the same id because that causes issues, so OCaml parser id is `refmt-ml`.

However, I didn't want to show that name on the toolbar, so using `displayName` instead, which doesn't pose issues if duplicated.

website/src/components/getFocusPath.js
```diff
@@ -34,6 +34,8 @@ export default function getFocusPath(node, pos, parser, seen = new Set()) {
     if (value && typeof value === 'object' && !seen.has(value)) {
       let childPath = getFocusPath(value, pos, parser, seen);
       if (childPath.length > 0) {
+        // if current wasn't added, add it now
+        childPath = range ? childPath : [node].concat(childPath);
```
jchavarri: In OCaml AST case, not all nodes throughout the tree have location information.

This change allows to include a child in the path even if the call to `nodeToRange` above didn't return a range, but only _if we know there are deeper descendants in the path_.

This allows to provide a more accurate node selection in many cases.

website/src/containers/CodeEditorContainer.js
```diff
@@ -7,7 +7,7 @@ function mapStateToProps(state) {
   return {
     keyMap: getKeyMap(state),
     value: getCode(state),
-    mode: getParser(state).category.id,
+    mode: getParser(state).category.editorMode || getParser(state).category.id,
```
jchavarri: There's no syntax highlight for Reason in codemirror... but `rust` gets us 90% of the way. In order to keep the `reason` id but be able to pass a different mode to codemirror, `editorMode` was added to category.

website/src/parsers/reason/index.js
```diff
@@ -0,0 +1,8 @@
+// Rust syntax highlight works well enough with Reason
+import 'codemirror/mode/rust/rust';
+
+export const id = 'reason';
+export const editorMode = 'rust';
```
jchavarri: No `reason` mode in codemirror yet.

website/src/components/Toolbar.js
fkling: That's the whole purpose of `displayName`. Guess I forgot to use it 😆

website/src/components/getFocusPath.js
fkling: FWIW I'm actually in the process of removing this code and replace with something that uses https://github.com/fkling/astexplorer/blob/540c2311c51d451fd1d97e03b2589ebd4606ee6e/website/src/core/TreeAdapter.js#L26-L81 instead.

fkling:
Amazing, thank you! Looks good to me.

Eventually I want to be able to build every parser in isolation, so some of the code may have to duplicated in future (or at least refactored), but we can worry about this when it happens.

I'll see that I get this merged in the next few days.

website/src/containers/CodeEditorContainer.js
thangngoc89: @jchavarri see https://github.com/Sketch-sh/sketch-sh/blob/master/client/src_editor/reason-mode.js for hljs

jchavarri: @thangngoc89 Thanks! I added this in https://github.com/fkling/astexplorer/pull/400/commits/88e9675a15a2627b8dbe73df6cfb7f76bc29a07b.

As an aside, do you know what it'd require to make this a proper codemirror mode package?

website/src/components/getFocusPath.js
jchavarri: Cool! it seems (from a very quick look haha) that this code would continue to cover this case, right?

website/src/containers/CodeEditorContainer.js
thangngoc89: @jchavarri I have no ideas on how to make it a package. I asked codemirror's author for upstreaming that but they said they aren't accepting any new mode. If you are willing to maintain a npm package for it, please do

website/src/components/getFocusPath.js
fkling: Yep!

fkling:
It's live!

PR #401, opened 2019-05-04
Refactor autofocus and highlight logic

fkling:
To further eliminate the need to pass the parser object to visualization
components, the "focus" path is now computed on the fly focus
computation.

That means that every node computes itself whether it should be open and
whether it should be highlighted. This is possible because the tree
adapter can answer the question whether a node contains children in
cursor range or not.

This commit also changes the behavior for nodes that have been opened
via focus: If nodes loose focus, they used to stay open. Now, they will
close as long as they haven't been explicitly opened.


fkling:
Merged starting at 83696c133cb89b7dd07a07de236c99d9083468fa.

PR #406, opened 2019-05-22
deps: Update @typescript-eslint/parser and typescript

bradzacher:
Need to upgrade to support new syntax that has been added (such as `type T = readonly number[]`


fkling:
Sorry for the long radio silence. I was on vacation and I'm currently in process to move my family across continents.

I assume new versions have been released in the meantime. We can update later.

PR #414, opened 2019-06-18
feature: add putout

coderaiser:
Thank you for such a useful tool as `AST Explorer` :), it helped me a lot to learn what `JavaScript AST` consists from and how to interact with. Also it helps me a lot to work on Pluggable Code Transformer [Putout](https://github.com/coderaiser/putout).
This is something like `eslint` but it works not with formatting but with code, for example it can help you to replace such code:

```diff
function merge(a) {
-   return Object.assign({}, a, {
-       hello: 'world'
-   });
+   return {
+       ...a,
+       hello: 'world'
+   };
};
```

It consists of  `babel` and `recast` and this provides a lot flexibility with AST-manipulation and transform only code that matters, keeping untouched everything else.
And the most important and powerful thing is it plugins. There is no built-in rules, everything is a plugin.

`Putout` uses similar but little bit different then `babel` plugin structure:

```js
const {
  types,
} = require('putout');

// when find node that can be fixed tell about it 
module.exports.report = ({path}) => `identifier ${path.node.name} should be reversed`;

// this part will run only when `putout --fix` will be used
module.exports.fix = ({path}) => {
  path.node.name = path.node.name.split('').reverse().join('');
};

// this part used for search
module.exports.find = (ast, {push, traverse}) => {
  traverse(ast, {
    Identifier(path) {
      push({path})
    }
  });
};
```

Would be great if `putout` plugins could be created from https://astexplorer.net, this would be very useful :), for me and `putout` plugin developers.


eventualbuddha:
Would it make sense to make putout plugins also serve as babel plugins? Then you could use them as-is.

coderaiser:
@eventualbuddha, unfortunately no, because babel plugins only `transform`, and not `report`.
`Putout` can `find`, `fix` and `report`, this is a thing which is similar in `putout` and `eslint`(but `putout` has a more simple `api` with less nested objects).

Anyways `putout` supports [babel](https://github.com/coderaiser/putout#babel-plugins) and [jscodeshift](https://github.com/coderaiser/putout#jscodeshift) plugins in a slower and more limited mode then `native` plugins.

PR #415, opened 2019-07-03
upgrade ember-related dependencies

NullVoxPopuli:
before (on the currently deployed version):
image

after (locally):
image

(applies a bugfix for positional params, when replaced, would lose their spacing -- look at the difference in the bottom-right)


fkling:
Sorry for the long radio silence. I was on vacation and I'm currently in process to move my family across continents.

PR #419, opened 2019-07-19
flow 0.103

gkz:
Update to Flow 0.103. Add option (default off) to parse experimental enums. 

Tested through `yarn run build && yarn run start`, checking out website locally


fkling:
I think the `package.json` file needs to be updated as well, but I can do that. Thank you!

PR #424, opened 2019-07-25
Update @angular/compiler to latest stable version

SerkanSipahi:


fkling:
Are there any API changes or additional options in the new versions, that should be added to astexplorer?

fkling:
Based on the changelog I guess not.

PR #437, opened 2019-09-05
Fix overflowing toolbar menu

pcr910303:
This PR fixes the overflowing toolbar menu on smaller screens.

Before: 
image

After: 
image


fkling:
Thank you, this is great! I tweaked a little bit in a follow up commit so that it looks better in other environments.

PR #439, opened 2019-09-14
feature: add regjsparser

JLHwung:
In this PR we add regjsparser support to astexplorer.

Background: regjsparser is used by a popular library `regexpu-core`, which transpiles unicode related regular expressions to ES5. `@babel/preset-env` replies on `regexpu-core` for unicode regex features.

When we are investigating babel issues about unsupported regex, I realize it would be convenient to browse the ast generated by regjsparser, isolate the problem and to see if it is on the parser side or the transformer side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ The AST explorer provides following code parsers:
   - [php-parser][]
 - Regular Expressions:
   - [regexp-tree][]
+  - [regjsparser][]
 - Scala
   - [Scalameta][]
 - Solidity:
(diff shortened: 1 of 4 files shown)


JLHwung:
ping @fkling , anything I can help to move forward?

fkling:
Sorry @JLHwung! I'm just getting started investing more time into astexplorer again. I'll take a look in the next few days!

fkling:
Thank you and sorry for the long wait. I will update the PR once it's pushed to production.

Issue #438, opened 2019-09-11
Typescript isn't a parser for recast


dangreenisrael:
**Describe the bug**

Typescript is missing from the recast parser options

**To Reproduce**
1. Go to 'recast settings'
2. Click on 'parset'
3. See that "typescript" is not an option

**Expected behavior**
Typescript should be an option

**Screenshots**
<img width="390" alt="Screen Shot 2019-09-11 at 2 06 39 PM" src="https://user-images.githubusercontent.com/4509692/64723129-e517b180-d49d-11e9-8a43-11f0b996b41d.png">

**Browser (please complete the following information):**
 - OS: Mac OS Mojave
 - Browser: Chrome
 - Version: 76

**astexplorer settings:**
- Selected parser: n/a
- Selected transformer (if applicable): jscodeshift
- Contents of the local storage key `explorerSettingsV1` (code can be removed if you don't want it to be public)
```json
{
    "showTransformPanel": false,
    "parserSettings": {
        "recast": {
            "tolerant": false,
            "range": true,
            "parser": "esprima",
            "flow": {
                "enums": false,
                "esproposal_class_instance_fields": true,
                "esproposal_class_static_fields": true,
                "esproposal_decorators": true,
                "esproposal_export_star_as": true,
                "esproposal_optional_chaining": true,
                "esproposal_nullish_coalescing": true,
                "tokens": false,
                "types": true
            },
            "babylon6": {
                "sourceType": "module",
                "allowImportExportEverywhere": false,
                "allowReturnOutsideFunction": false,
                "tokens": false,
                "plugins": ["asyncGenerators", "classConstructorCall", "classProperties", "decorators", "doExpressions", "exportExtensions", "flow", "functionSent", "functionBind", "jsx", "objectRestSpread", "dynamicImport"]
            },
            "babylon7": {
                "sourceType": "module",
                "allowImportExportEverywhere": true,
                "allowReturnOutsideFunction": true,
                "ranges": true,
                "tokens": true,
                "plugins": ["asyncGenerators", "classProperties", "decorators", "doExpressions", "exportExtensions", "flow", "functionSent", "functionBind", "jsx", "objectRestSpread", "dynamicImport", "numericSeparator", "optionalChaining", "optionalCatchBinding", "typescript"]
            }
        }
    },
    "parserPerCategory": {
        "javascript": "recast",
        "css": "postcss"
    },
    "workbench": {
        "parser": "recast",
        "keyMap": "default",
        "transform": {
[...]

PR #441, opened 2019-09-15
Issue #438 - Add typescript as a parser for recast

dangreenisrael:
This adds typescript support for recast.

This follows the example given in the recast docs

```javascript
const tsAst = recast.parse(source, {
  parser: require("recast/parsers/typescript")
});
```
https://github.com/benjamn/recast#using-a-different-parser

### DIFF
--- a/website/src/parsers/js/recast.js
+++ b/website/src/parsers/js/recast.js
@@ -19,8 +19,8 @@ export default {
 
   loadParser(callback) {
     require(
-      ['recast', 'babel5', 'babylon6', 'babylon7', 'flow-parser'],
-      (recast, babelCore, babylon6, babylon7, flow) => {
+      ['recast', 'babel5', 'babylon6', 'babylon7', 'flow-parser', 'recast/parsers/typescript'],
+      (recast, babelCore, babylon6, babylon7, flow, typescript) => {
         callback({
           recast,
           parsers: {
@@ -28,6 +28,7 @@ export default {
             babylon6,
             babylon7,
             flow,
+            typescript,
           },
         });
       }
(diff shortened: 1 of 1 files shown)


dangreenisrael:
@fkling I was hoping you may have some insight as to why this issue is happening:

The parser works, and I can click through the AST, but as soon as a I click a node a `tokens` node pops up, I click that and an error boundary gets thrown. The trace the console is not super helpful.


Screen Shot 2019-09-14 at 10 45 53 PM
Screen Shot 2019-09-14 at 10 45 58 PM
Screen Shot 2019-09-14 at 10 46 02 PM
Screen Shot 2019-09-14 at 8 56 50 PM

fkling:
Thanks for taking a stab at it. I will merge the PR locally and investigate the error.

However, if we have to actually `require` the typescript parser (as it appears to be the case, I thought passing a string would suffice, but I guess that changed), then we have to do it slightly differently anyway. The parser would have to be loaded further in the `loadParser` method, just like the other modules.

dangreenisrael:
Thanks @fkling. Please let me know if there is anything I can do to be helpful.

website/src/parsers/js/recast.js
```diff
@@ -68,7 +68,9 @@ export default {
       case 'babel5':
         options.parser = parsers[options.parser];
         break;
-      case 'esprima':
+      case 'typescript':
+        options.parser = require('recast/parsers/typescript')
```
fkling: This needs to be moved into the `loadParser` method, otherwise the module is going to be bundled with the core instead of being lazy loaded when recast is selected.

dangreenisrael:
Ahhh, awesome. I was just about to open another issue because I noticed a similar behaviour in prod.

dangreenisrael:
Looks like it's working now.

Screen Shot 2019-09-21 at 5 32 18 PM

dangreenisrael:
Thank you!

PR #457, opened 2019-11-09
WIP: [parsers] Add tree-sitter POC

forivall:
At the very least, this is waiting on https://github.com/tree-sitter/tree-sitter/pull/481 to be merged and released, so the patch-package hackage can be removed.

Very POC, requesting feedback. As tree-sitter's parsing internals are written in C, this requires some changes to get the WASM binary to load, so feedback on that in particular would be appreciated.


fkling:
So, am I correctly understanding that this is the JavaScript parser generated by tree sitter?
I see that it supports various languages. Does that mean we could also add a parser for e.g. Bash?

forivall:
Exactly, yes, this is the JavaScript parser that tree-sitter generates (from tree-sitter-javascript). I wanted to see what the structure its JavaScript AST looked like and see if it's ESTree or not, and the best way i thought of was to submit it here, then other people can look at how tree-sitter works!

For the incremental stuff to work fully, it looks like we have to notify the tree-sitter internals about the edits (see https://github.com/tree-sitter/tree-sitter/blob/9ce3a5/docs/assets/js/playground.js#L105 ), but getting that to work would take additional changes to astexplorer's saga data model, and i just didn't have time for that then.

PR #465, opened 2019-11-26
Pug

### LINKED ISSUE #464: Add support for Pug parser
PR on its way :slightly_smiling_face:

vilarfg:
fixes #464

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ The AST explorer provides following code parsers:
   - [mdxhast][]
 - PHP
   - [php-parser][]
+- [Pug][]
 - Regular Expressions:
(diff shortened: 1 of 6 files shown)


fkling:
"Fixed" the eslint issues :wink: Thank you! Will update once it's deployed to production.

fkling:
Does the AST not contain any location information that could be used to highlight nodes?

vilarfg:
It only contains start line and column info for nodes. No start, end or
range info given.
I would need access to the whole string to compute `[start, end]`, but even
then it would be cumbersome (including computing whole length of node,
which is not given either).
So, guess not... 😭

PR #475, opened 2019-12-27
Add .editorconfig

junoatscroll:
My previous PR failed in travis because my editor used the defaults (double quotes, etc). This adds a .prettierrc file which conforms afaict to the project's existing code and lint settings.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*.js]
+quote_type = single


G-Rath:
I don't think theres a lot of value in adding a `.prettierrc` w/o actually using prettier in the repo, as it would be very easy for them to become out of sync.

Instead, I think it would make more sense to add an `.editorconfig` w/ `quote_type` set to `single` (which Prettier should respect iirc).

junoatscroll:
Fair enough! Updated with @G-Rath 's suggestion.

fkling:
I don't have a use for this personally, so I won't be able to maintain this properly, but if it helps others then I'm happy ;)

PRINCEBTC1:
> My previous PR failed in travis because my editor used the defaults (double quotes, etc). This adds a .prettierrc file which conforms afaict to the project's existing code and lint settings.

0x8179FE8F7f7D95cd02a3d87eC7FeAfcf2078325e

Issue #478, opened 2020-01-26
Whitespace in values is collapsed


Turbo87:
**Describe the bug**

When using the Handlebars parser and looking at the value of a `TextNode` (`chars` attribute), I would expect the value to display whitespace characters correctly, but due to the default whitespace collapsing of the browser they are only shown as a single space character instead, which can be quite confusing.

**To Reproduce**
1. Go to https://astexplorer.net/#/gist/179c0fc002d69cbac732c71e00e7a5bc/1d3aed24698ba0cf5f3fedd6025e61ee7f29eeb4
2. Click on 'bar'
3. See `" bar\n"` in `chars` of the selected `TextNode`

**Expected behavior**
like in JSON mode, I would expect to see `"      bar\n"` instead. (thanks GitHub for collapsing the whitespace too 😅)

```
"      bar\n"
```

**Browser (please complete the following information):**
 - OS: macOS
 - Browser: Chrome, Firefox, Safari, ...

**astexplorer settings:**
- Selected parser: Handlebars

**Additional context**

I would suggest to put `white-space: pre-wrap;` on the `.value` CSS class to resolve this issue.

jedwards1211:
Same goes for `JSXText` nodes

PR #479, opened 2020-01-26
CSS: Apply `white-space: pre-wrap;` to values

Turbo87:
Resolves https://github.com/fkling/astexplorer/issues/478

### DIFF
--- a/website/css/style.css
+++ b/website/css/style.css
@@ -352,6 +352,10 @@ li.entry {
   background-color: rgba(255,240,6,0.4);
 }
 
+.entry > .value {
+  white-space: pre-wrap;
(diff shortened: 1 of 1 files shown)


Turbo87:
@fkling any thoughts on this?

fkling:
Sorry for the late reply (needed a break and was busy the last couple of days).

My thoughts are: Why didn't anyone notice this before?? :wink:

jedwards1211:
Thanks y'all!  Haha I only noticed it when I started looking into splitting JSXText nodes, so that I could do various wrapping transforms on only the text and other JSX children the user had selected.

PR #481, opened 2020-02-14
enable nullishCoalescingOperator for babylon7 by default

antonk52:
I would like to propose enabling `nullishCoalescingOperator` for babylon7 by default, given that it is enabled for `babel-plugin-macros`, `flow` parser. It also works in typescript parser out of the box and it is now works in the [latest chrome and firefox versions](https://caniuse.com/#feat=mdn-javascript_operators_nullish_coalescing). I think it is a good idea for this plugin to be enabled by default.

### DIFF
--- a/website/src/parsers/js/babylon7.js
+++ b/website/src/parsers/js/babylon7.js
@@ -56,6 +56,7 @@ export const defaultOptions = {
     'jsx',
     'objectRestSpread',
     'dynamicImport',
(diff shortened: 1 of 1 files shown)


fkling:
Seems reasonable. TBH I don't know/remember why some options are enabled and some are not. Maybe just an oversight.

IMO as long as settings are not conflicting with others or specific behavior they should be enabled by default.

PR #485, opened 2020-02-25
Update angular/compiler to 9.0.2

irustm:


fkling:
Was updated in another PR. Thank you though!

PR #493, opened 2020-03-21
Add thrift IDL

pd4d10:
Parser here: https://github.com/creditkarma/thrift-parser


fkling:
Thank you! I will make a small change to load the package asynchronously.

PR #497, opened 2020-04-17
feat: make "Vue" a standalone category and add several alternative parsers

sodatea:
Currently, there are 3 different parsers available for parsing `.vue` files.

Also, the `.vue` file format differs enough from the HTML format. For example, it does not support the `doctype` declaration.

So I think it makes more sense to have a standalone "Vue" category.


website/src/parsers/vue/vue-eslint-parser.js
```diff
@@ -0,0 +1,45 @@
...
+    return key === 'children';
+  },
+
+  getNodeName(node) {
+    return node.tag;
+  },
+
+  getDefaultOptions() {
+    return {
+      sourceType: 'module'
```
IWANABETHATGUY:
```suggestion
      sourceType: 'module',
```

website/src/parsers/vue/vue-eslint-parser.js
```diff
@@ -0,0 +1,45 @@
...
+    };
+  },
+
+  _ignoredProperties: new Set(['parent', 'tokens'])
```
IWANABETHATGUY:
```suggestion
  _ignoredProperties: new Set(['parent', 'tokens']),
```

sodatea:
I've also deployed a preview site on Netlify: https://frosty-hodgkin-d6766b.netlify.app/

fkling:
Seems very reasonable to me, thank you!

PR #500, opened 2020-05-09
refurbish babel 7.9 support

JLHwung:
This PR refurbishes babel 7.9 support:

- Added more available proposals and language extensions
- Added more parser options
- Removed stage-4 proposals which are always enabled in `@babel/parser`
- Added stage-3 proposals to default options


fkling:
Thank you!

PR #504, opened 2020-05-22
Add Hermes parser for JavaScript

motiz88:
(EDIT: Updated June 2021.)

Adds the [Hermes engine](https://github.com/facebook/hermes)'s parser into AST Explorer. Try it at [astexplorer-hermes.now.sh](https://astexplorer-hermes.now.sh).

screenshot of AST Explorer with the Hermes parser

Implementation notes:
1. This uses the official [hermes-parser](https://www.npmjs.com/package/hermes-parser) npm package and supports all of its options (except `sourceFilename`).
2. We wrap the parser in a Web Worker, because some browsers will refuse to synchronously load a WebAssembly module on the main thread when it's as big as the one `hermes-parser` uses. As a side benefit, this is also nice for UI perf.
3. There is some subtlety to how errors are reported because of the worker architecture - sending `Error` instances over `postMessage` behaves differently across browsers so I've ditched that. The current solution is portable (tested in Chrome, Firefox and Safari) and preserves the error location data for use in the editor.
4. I've snuck in a tiny CSS change so that errors are displayed in a monospace font with whitespace preserved. This improves error output for parsers (like `hermes-parser`) that include code frames in their error messages.
5. Hermes and `hermes-parser` are versioned independently, so to prevent confusion I used `hermes-parser` as the display name even though the word "parser" is redundant in context.

### DIFF
--- a/website/src/parsers/js/hermes.js
+++ b/website/src/parsers/js/hermes.js
@@ -1,0 +1,50 @@
+import defaultParserInterface from './utils/defaultESTreeParserInterface';
+// NOTE: We load the hermes-parser package in a worker and not directly, because
+// it violates the typical limit on the size of a WebAssembly module that can be
+// compiled synchronously on the main thread.
+import HermesWorkerClient from './hermes/HermesWorkerClient';
+import pkg from 'hermes-parser/package.json';
+
+export const defaultOptions = {
+  sourceType: 'unambiguous',
+  flow: 'detect',
+  allowReturnOutsideFunction: false,
+  babel: false,
+  tokens: false,
+};
+
+export const parserSettingsConfiguration = {
+  fields: [
+    ['sourceType', ['unambiguous', 'module', 'script']],
+    ['flow', ['detect', 'all']],
+    'allowReturnOutsideFunction',
+    'babel',
+    'tokens',
+  ],
+};
+
+export default {
+  ...defaultParserInterface,
+
+  id: 'hermes',
+  displayName: pkg.name,
+  version: pkg.version,
+  homepage: pkg.homepage || 'https://hermesengine.dev/',
+  locationProps: new Set(['range', 'loc']),
+
+  loadParser(callback) {
+    callback(new HermesWorkerClient());
+  },
+
+  async parse(hermes, code, options) {
+    return await hermes.parse(code, options);
+  },
+
+  getDefaultOptions() {
+    return defaultOptions;
+  },
+
+  _getSettingsConfiguration() {
+    return parserSettingsConfiguration;
+  },
+};
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ The AST explorer provides following code parsers:
   - [esformatter][]
   - [esprima][]
   - [flow-parser][]
+  - [hermes-parser][]
   - [meriyah][]
   - [recast][]
   - [seafox][]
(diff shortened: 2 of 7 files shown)


website/src/parsers/js/hermes.js
```diff
@@ -0,0 +1,44 @@
...
+  version: 'N/A',
+  homepage: 'https://hermesengine.dev/',
+  locationProps: new Set(['loc', 'range']),
+
+  loadParser(callback) {
+    callback(new HermesWorkerClient());
+  },
+
+  async parse(hermes, code, options) {
+    const hermesOutput = await hermes.run(['-dump-ast', '--pretty-json', '-dump-source-location', options.jsx ? '-parse-jsx' : ''].filter(Boolean), code);
```
RReverser, line 28: Does Hermes not provide a library-based API, only a binary?

website/src/parsers/js/hermes.js
```diff
@@ -0,0 +1,44 @@
...
+    try {
+      return JSON.parse(hermesOutput);
+    } catch (e) {
+      console.error(e);
```
RReverser, line 32: Console leftover? I don't think we do this in other plugins.

website/src/parsers/js/hermes.js
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error(hermesOutput);
```
RReverser, line 33: Why does this ignore original error object? Which one has actual detailed information?

website/src/parsers/js/hermes.js
motiz88: I believe we're in the process of figuring out APIs for various similar integrations, not sure if there's anything very concrete yet that would fit here. (cc @tmikov)

website/src/parsers/js/hermes.js
motiz88: Yeah I can probably delete this. See rationale for having it in the first place in the comment below.

website/src/parsers/js/hermes.js
RReverser: Ah I see. Yeah, looks like currently this integration doesn't provide proper source location for errors:
image

website/src/parsers/js/hermes.js
tmikov: I think ideally we should do something like https://github.com/facebook/hermes/blob/master/tools/emhermesc/HermesCompiler.js , but linking only the parser instead of the whole compiler and VM.

Even more ideally :-), we can construct the JS AST by calling JS callbacks from WASM instead of going through JSON, though wonder how much faster it would be. It shouldn't matter for this particular use case.

We really should settle these interfaces soon, before we publish an NPM and unblock this PR (in some form).

fkling:
This is great! At the very least it introduced me to `worker-loader`, which could be very useful for other parsers as well.
I guess I just wait until you think this is ready?

motiz88:
@fkling Strictly speaking, I was being cautious in saying this is blocked. If you're interested and don't mind the quirks of this integration too much (namely: checking in `hermes.js` and `hermes.js.mem` under `website/public`, not having structured error output), then IMO there's nothing stopping us from merging this and upgrading to the "proper" npm version later when that's available.

motiz88:
Hey @fkling, it's been over a year but I've finally gotten around to making this shippable. There's now an official `hermes-parser` package and it works nicely. When I tried this a few months ago, something in the WebAssembly build was not browser-compatible, but that seems to have been resolved.

website/src/parsers/js/hermes.js
```diff
@@ -0,0 +1,50 @@
+import defaultParserInterface from './utils/defaultESTreeParserInterface';
+// NOTE: We load the hermes-parser package in a worker and not directly, because
+// it violates the typical limit on the size of a WebAssembly module that can be
```
RReverser, line 3: Why doesn't it use async compilation?

motiz88: It's primarily intended for use in Node, typically as a drop-in replacement for other parsers that can be `require`d synchronously (e.g. `@babel/parser`), and async compilation would leak into the API and break these use cases. We could probably add a browser-specific entry point to the package that has an async API, but I'm not convinced that's desirable.

Caveat: I don't represent the Hermes team - this is based on my personal understanding. @avp might be able to shed more light on this.

fkling:
Very nice... my re-architecture effort of astexplorer moves every parser to workers anyway, so we will probably able to unify things again at some point.

PR #506, opened 2020-05-22
JavaScript, Babylon7, support selecting pipeline proposal

xixixao:
Supports selecting whichever pipeline proposal.

I'm gonna add "hack" (hopefully after adding it to Babel) later.

Before this commit:
Screen Shot 2020-05-22 at 16 32 43

After this commit:
Screen Shot 2020-05-22 at 16 32 30


fkling:
Thank you, this is a good idea!

Issue #508, opened 2020-05-27
[Feature request]: Add babel-typescript


SukantPal:
Babel can parse typescript if you add "typescript" to the list of plugins. If you can add this parser option, it would be really great. It shouldn't take a lot of effort.

fkling:
If you can provide a PR, I'd be happy to review it! :slightly_smiling_face:

fkling:
Typescript can be selected as an option for the babel parser.

PR #509, opened 2020-05-30
babel-with-typescript

SukantPal:
#508


fkling:
After revisiting this, I'm not sure why this is needed. You can already select `typescript` in the options of the existing parser... can you explain?

SukantPal:
@fkling The "typescript" parser's AST is different from that of Babel.

sibelius:
image

```jsx
const AppType = new GraphQLObjectType<IApp, GraphQLContext>({
  name: 'App',
  description: 'Represents a App',
  fields: () => ({
    id: globalIdField('App'),
    createdAt: {
      type: GraphQLString,
      resolve: (obj) => (obj.createdAt ? obj.createdAt.toISOString() : null),
    },
  }),
  interfaces: () => [NodeInterface],
});
```

this fails to parse using normal babel

if I switch to typescript it works, but it does not help if I'm working in codemods with jscodeshift that uses babel parser instead of typescript parser

sibelius:
if you check prettier parsers https://prettier.io/docs/en/options.html#parser

they have a `babel`, `babel-flow` and `babel-ts`

fkling:
I still don't understand why we need a separate entry for this. What is the problem with opening the settings and enabling the typescript option?

sibelius:
I've missed the settings, it works great tks

SukantPal:
@fkling I got it now. The "settings" button doesn't look separate from the parsers menu. This is a UI/UX/usability bug, I would say. I can close this PR, I guess then.

Screen Shot 2020-08-04 at 6 24 24 PM

fkling:
@SukantPal Agreed that it's not the most obvious of places (but it's been like this for a long time). I'm working on a new design.

yuzai:
this ui bug is confusing, I did't find the settings button until I saw this issue

PR #512, opened 2020-06-13
Rename babylon7 to @babel/parser

### LINKED ISSUE #502: Babylon should be renamed as @babel/parser
**Describe the bug**

Babylon is the old name for babel parser. It is no more used in documentation also. So it is confusing for the first timers to clearly know which parser to choose for babel when in astexplorer.  So I would recommend to rename it to `@babel/parser` to avoid any confusion.

**Expected behavior**
It should `@babel/parser` in parsers drop down.

**Additional context**
I would like to raise a PR.

AjayPoshak:
This PR fixes #502

### DIFF
--- a/website/src/parsers/js/babylon7.js
+++ b/website/src/parsers/js/babylon7.js
@@ -83,7 +83,7 @@ export default {
   ...defaultParserInterface,
 
   id: ID,
-  displayName: ID,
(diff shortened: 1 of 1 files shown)


fkling:
Thank you! This was the right move. If we also changed the id, then existing shared snippets wouldn't load correctly.

PR #513, opened 2020-06-13
fix: Added vue-compiler-core parsing options

privatenumber:
## Problem
- Parsing error'd on [native void elements](https://github.com/vuejs/vue-next/blob/master/packages/shared/src/domTagConfig.ts#L30) (eg. `` or `<br>`)

```html
<template>
  <img src="">
</template>
```

- Didn't correctly identify the `tagType` on non-[native elements](https://github.com/vuejs/vue-next/blob/master/packages/shared/src/domTagConfig.ts#L6-L28)

- Doesn't correctly parse JS in `script` tag:
```html
<template>
	<div>hello</div>
</template>

<script>
a < b;
export {};
</script>
```

## Changes
- Used `@vue/compiler-dom` instead of `@vue/compiler-core`, which pre-configures `baseParse` to be able to parse DOM SFCs

cc @sodatea (https://github.com/fkling/astexplorer/pull/497)


privatenumber:
Found that it fails on some `script` cases. For example:
```
<template>
	<div>Test</div>
</template>

<script>
export default {
	created() {
      	const version = 1;
		this.versions = data.filter(v => version < v));
	},
};
</script>
```
And that importing from `parse` from `@vue/compiler-dom` comes with all the right configurations necessary to parse a SFC.

Will update PR.

sodatea:
What about making it an option?

After all, there are more than one compilers extended from `@vue/compiler-core` for Vue 3, such as `@vue/compiler-ssr`.

Users may need to debug either one of them, or write their own compiler based on `@vue/compiler-core`.

privatenumber:
Good point! I'll look into it

privatenumber:
I can't tell how actively maintained this project is so I'm going to hold off until https://github.com/fkling/astexplorer/pull/515 and https://github.com/fkling/astexplorer/pull/514 are reviewed/merged before I spend more time on something that might not be merged

PR #514, opened 2020-06-13
chore: disable package-lock

privatenumber:
I assumed `package-lock.json` wasn't meant to be checked in, seeing as it wasn't already in the repo.

Disabling it at a repo level so it doesn't get generated in the first place.


fkling:
Thank you, that seems reasonable though this project uses `yarn` anyway.

PR #474, opened 2019-12-27
Add java parser - fixes #473

junoatscroll:
I followed the instructions in readme about adding a new parser- pretty painless! Please let me know if there are any changes you'd like or other review comments.

PR #516, opened 2020-06-15
Juno/add java parser (resolved merge conflicts from PR 474)

aceslick911:
Hi,
I've forked from this PR: https://github.com/fkling/astexplorer/pull/474 and merged your latest master in to it so we can merge the java parser finally.

Changes:
> Merged fkling/astexplorer master into juno/add-java-parser (Resolved merge conflicts on package.json and yarn.lock)

All credit goes to @junoatscroll


website/src/parsers/java/java-parser.js
```diff
@@ -0,0 +1,51 @@
...
+  typeProps: new Set(['name']),
+
+  loadParser(callback) {
+    require(['java-parser'], callback);
+  },
+
+  parse(parser, code) {
+    const cst = parser.parse(code);
+
+    const clone = JSON.parse(JSON.stringify(cst));
```
fkling, line 30: Thank you for doing this! I still don't know why this line is necessary.... any ideas?

website/package.json
```diff
@@ -72,6 +72,7 @@
     "babylon6": "npm:babylon@^6",
     "babylon7": "npm:@babel/parser@^7",
     "cherow": "^1.6.8",
+    "chevrotain": "^7.0.1",
```
fkling, line 75: If this is installed as a dependency of `java-parser`, this shouldn't really be here...

aceslick911: Removed 👍

website/src/parsers/java/java-parser.js
aceslick911: This was possible debug code by the author. I've removed it.

aceslick911:
Removed debug code in java-parser.js
```javascript
  parse(parser, code) {
    const cst = parser.parse(code);

    const clone = JSON.parse(JSON.stringify(cst));

    return clone;
  },
```

-> 

```javascript
  parse(parser, code) {
    return parser.parse(code);
  },
```

junoatscroll:
Thank you for taking the lead on this!
The json stringify/parse was a super hacky way of cleaning up some of the AST data structures that had cycles and were making ASTexplorer unhappy. YMMV. A better approach would be to writer better filters in the parser config.

Unfortunately, I don't have capacity to work further on this at the moment.

fkling:
@junoatscroll 

> The json stringify/parse was a super hacky way of cleaning up some of the AST data structures that had cycles and were making ASTexplorer unhappy.

Are you sure that was the actual problem? Because `JSON.stringify` won't actually work if you have cyclic references... but anyway, this can also be solved by adding "offenders" to `_ignoredProperties`.

I will work on getting a new deployment process up an running first and then do a double check whether I encounter any issues with this parser.

PR #521, opened 2020-07-20
Close #482. Add Filbert for Python language support

### LINKED ISSUE #482: [feature] Python typed-ast
add https://pypi.org/project/typed-ast/ to the explorer

Glavin001:
Close #482 

Uses https://github.com/codecombat/filbert

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -88,6 +88,7 @@
     "eslint4": "npm:eslint@^4",
     "espree": "^6.2.1",
     "esprima": "^4.0.1",
+    "filbert": "^0.1.20",
     "flow-parser": "^0.122.0",
     "font-awesome": "^4.5.0",
     "glsl-parser": "^2.0.0",
--- a/website/src/parsers/python/codeExample.txt
+++ b/website/src/parsers/python/codeExample.txt
@@ -1,0 +1,20 @@
+print("Hello world")
+
+num1 = 1.5
+num2 = 6.3
+
+# Add two numbers
+sum = num1 + num2
+
+# Display the sum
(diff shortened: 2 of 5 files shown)


fkling:
I wonder how active this project still is, but better than nothing (who am I to judge...)

I recently migrated astexplorer to a different server and have to figure out a new deployment process before I can make any updates. Sorry.

Glavin001:
> I recently migrated astexplorer to a different server 

I thought astexplorer is client side only. What is the server used for?

> new deployment process 

Would something like https://vercel.com/ or https://www.netlify.com/ work? I'd be up for helping out, if you could explain more about the constraints and setup you have in mind.

fkling:
> I thought astexplorer is client side only. What is the server used for?

astexplorer used to use Parse as storage backend, which got sunset. In order to preserve existing links, a copy of the data is stored on the server and served from it.

The server also holds the secret for storing snippets as gists, i.e. saving and loading snippets goes through the server.

You can find that code here: https://github.com/fkling/astexplorer/tree/7d6100272ca93fc096dbd505ddd8a59fe09c41be/server

Eventually I also want the server to periodically rebuild parsers to update them automatically.

> Would something like https://vercel.com/ or https://www.netlify.com/ work? I'd be up for helping out, if you could explain more about the constraints and setup you have in mind.

Thank you! Probably not with the way things are currently build. I already setup CI/CD pipelines in the GitLab mirror and that has been working well so far. I think I just need to setup some scripts to sync the artifacts to the new server via `rsync`. But I'm happy to get back to you if that doesn't work out somehow.

PR #522, opened 2020-07-23
Support parsing and transform HTML with rehype

pinkiebala:
Hi, I'd like to add support of parsing HTML with [rehype](https://github.com/rehypejs/rehype) and got a question here.

How can I pass the options of the AST parser to the parser used in the transformer? [AST parser option](https://github.com/pinkiebala/astexplorer/blob/9dfcaaa9fbdeb4e9f26e11e99da6c0c1c044929c/website/src/parsers/html/rehype.js#L35), [transformer parser option](https://github.com/pinkiebala/astexplorer/blob/9dfcaaa9fbdeb4e9f26e11e99da6c0c1c044929c/website/src/parsers/html/transformers/unified/index.js#L42). Or maybe reuse the tree from the former parser?


fkling:
> How can I pass the options of the AST parser to the parser used in the transformer? AST parser option, transformer parser option. Or maybe reuse the tree from the former parser?

None of this is currently possible, but I'm trying to figure it out. Essentially I think that instead of transformers being something that you enable in *addition* to a parser, a transformer has to expose its own `parse` method and has to be selectable by itself.

---

I'm recently migrated astexplorer to a different server and have to figure out a new deployment process before I can make any updates. Sorry.

pinkiebala:
> Essentially I think that instead of transformers being something that you enable in addition to a parser, a transformer has to expose its own parse method and has to be selectable by itself.

That's a good point! Thank you for your reply. I'm looking forward to the updates.

Btw, This is a great tool even just create my own version on the localhost, thanks! :smile:

PR #533, opened 2020-09-09
Add new parser: @babel/eslint-parser

edwining01:
Support @babel/eslint-parser@^7.11.5 as parser


fkling:
Thanks. Looks like once babel v8 is released we can remove `babel-eslint` in favor of this one.

PR #543, opened 2020-10-19
Fix avoid hiding empty array elements

bmeck:
This allows holes to be visible in code like:

```mjs
let foo = [1,,3];
let [a,,c] = foo;
```

This also means humans reading arrays can see the correct index while viewing.


fkling:
Thanks you! That makes a lot of sense. I never noticed that this is an issue.

PR #544, opened 2020-10-24
Update Rust parser to syn 1.0

dtolnay:
This PR pulls in the newest version of the Rust parser:

- A few syntax tree changes in the underlying Rust library between 0.15 and 1.0: https://github.com/dtolnay/syn/releases/tag/1.0.0#user-content-breaking-changes

- The corresponding PR to the npm package: https://github.com/RReverser/astexplorer-syn/pull/1

---

Screenshot

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -59,7 +59,7 @@
     "acorn-to-esprima": "^2.0.8",
     "astexplorer-go": "^1.0.0",
     "astexplorer-refmt": "^1.0.1",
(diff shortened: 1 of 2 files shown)


RReverser:
Just in case - note that the publishing is currently a manual process, so the website will be updated when @fkling has time to do the deploy.

fkling:
Should be live now (might need force refresh).

RReverser:
Awesome, thanks!

PR #545, opened 2020-10-26
update babel to update babel types

bmeck:
this fixes `t.arrayPattern([null])` to not throw errors


fkling:
Should be updated now! Thank you.

PR #548, opened 2020-11-06
feat: bump remark to 13

gengjiawen:


ChristianMurphy:
Thanks @gengjiawen!
A note, remark 13 removed the options currently showing in ast exporter.
`commonmark` became the new default, `gfm` is now a plugin, and `pedantic` is no longer supported.
I opened a follow up PR #550 that updates the settings to account for this.

PR #550, opened 2020-11-15
upgrade remark to version 13

ChristianMurphy:
remark 13 moves several parser options, like support for github flavored markdown into plugins.
To account for this, four of remark's most common parser plugins can now be enabled through settings.
In addition, remark is now included as a transformer, to allow testing new transforms/plugins.

alternative to #548 
closes #541 

/cc @wooorm

PR #551, opened 2020-11-20
include more complete remark transform example

ChristianMurphy:
addresses https://github.com/fkling/astexplorer/pull/550#discussion_r527588560

this adds an example transform that increases the depth of headings by 1, `# header` to `## header`, `## header` to `### header` etc.


wooorm:
Can we somehow add a package in ASTExplorer?
It would be a bit nicer if folks can use the common package `unist-util-visit` to walk the tree

wooorm:
semicolon use is inconsistent btw

ChristianMurphy:
> Can we somehow add a package in ASTExplorer?

In ASTExplorer yes.
Into examples, not that I have found (yet).
in examples both `import` and `require` are limited/unavailible.

fkling:
I’m not on my computer right now, but it’s possible to make additional packages available to transforms, it just needs to be preconfigured. There is another transform that does that I think.
I can provide more info when I’m at my computer.

fkling:
The second argument to `compileModule` allows you to specify which variables should be available in the transform. You can either do what the postcss transform does and [provide a fake `require` function](https://github.com/fkling/astexplorer/blob/860b77e178e2cff9fbb041a1b38527524dd3cd0a/website/src/parsers/css/transformers/postcss/index.js#L22-L32), or you can the tools available directly, [like babel-plugin-macros](https://github.com/fkling/astexplorer/blob/860b77e178e2cff9fbb041a1b38527524dd3cd0a/website/src/parsers/js/transformers/babel-plugin-macros/index.js#L24-L27).

ChristianMurphy:
Thanks for the tip @fkling! :bow: 
I've included three common utils in 152ae00 to simplify transform testing.

website/src/parsers/markdown/transformers/remark/index.js
```diff
@@ -1,5 +1,20 @@
 import compileModule from '../../../utils/compileModule';
 import pkg from 'remark/package.json';
+import unistUtilIs from 'unist-util-is';
+import unistUtilVisit from 'unist-util-visit';
+import unistUtilVisitParents from 'unist-util-visit-parents';
```
fkling, line 5: I'd prefer if those are also loaded dynamically, but if they are not too large then it's OK for now. You could create the `sandboxRequire` function in the `loadTransformer` method to make it available inside `transform`.

website/src/parsers/markdown/transformers/remark/codeExample.txt
```diff
@@ -1,5 +1,8 @@
+const visit = require("unist-util-visit");
```
fkling, line 1: I think we should include a comment here that explains which packages can be required.

website/src/parsers/markdown/transformers/remark/index.js
ChristianMurphy: addressed in 0e12f3c

website/src/parsers/markdown/transformers/remark/codeExample.txt
ChristianMurphy: addressed in 0e12f3c

fkling:
Awesome. It's live now :slightly_smiling_face:

PR #557, opened 2020-12-26
Switch to vercel/serve for yarn start

punkeel:
http-server does not send the proper MIME header for .wasm files, which prevents
development of WebAssembly-based parsers like swc. serve supports this use case,
and is edited by Vercel (previously Zeit) with 6.5k stars on GitHub.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -25,7 +25,7 @@
     "exports-loader": "^0.7.0",
     "file-loader": "^6.0.0",
     "html-webpack-plugin": "^3.2.0",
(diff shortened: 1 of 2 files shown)


gengjiawen:
Any thought on [serve](https://www.npmjs.com/package/serve), which have better popularity, also used by react and vue toolkit.

punkeel:
Good idea! I'll switch to that. It doesn't support live reloading, but that's fine … and it's used by plenty of devs already 👍

fkling:
Thank you!

Issue #554, opened 2020-12-15
feature request: support monkey


gengjiawen:
Site is in https://monkeylang.org/. It's pretty new language, pretty an amazing example demonstrate compiler internals. I wonder are you open to accept related PRs.

fkling:
I actually have those books :smile: 

Is there a JavaScript implementation though?

gengjiawen:
A wasm version generated by rust :)

PR #560, opened 2021-01-07
feat: add monkeylang

gengjiawen:
Fix https://github.com/fkling/astexplorer/issues/554

I don't know the webpack magic, but it just works.

image


gengjiawen:
@fkling PTAL.

fkling:
Sorry for the long wait! It could be that the way the WASM file is build needs to be updated at some point, but I will reach out if/when that's the case.
I'm working on a different way of building everything but it's unclear when that will be ready.

PR #563, opened 2021-02-02
Adds "tsx" as a parser within the recast parser

arnebech:
Adds a new option inside the Recast parser settings for a "tsx" parser.

Screen Shot 2021-02-02 at 8 18 36 AM

### DIFF
--- a/website/src/parsers/js/recast.js
+++ b/website/src/parsers/js/recast.js
@@ -19,8 +19,8 @@ export default {
 
   loadParser(callback) {
     require(
-      ['recast', 'babel5', 'babylon6', 'babylon7', 'flow-parser', 'recast/parsers/typescript'],
-      (recast, babelCore, babylon6, babylon7, flow, typescript) => {
+      ['recast', 'babel5', 'babylon6', 'babylon7', 'flow-parser', 'recast/parsers/typescript', './utils/recast-tsx-parser'],
+      (recast, babelCore, babylon6, babylon7, flow, typescript, tsx) => {
         callback({
           recast,
           parsers: {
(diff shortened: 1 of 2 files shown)


fkling:
How is that different from selecting the `babel7` parser and enabling the `typescript` and `jsx` options?

arnebech:
Maybe it's the same? But it wasn't obvious. In order for it to behave similarly you'd need to select "babel7", uncheck "flow", check "typescript" and then check "tokens", then it works the same/similarly. Especially having to check the "tokens" option wasn't really obvious to me. 

It would make it easier for users to have a tsx option at a high-level for recast, but totally understand if this is not the best way to go about it.

Here is the test code I used
```tsx
import React from 'react';

const Trends: React.FC = () => {
  return (
    <Container>
      <Icon />
    </Container>
  );
};

export default React.memo(Trends);
```

fkling:
I understand that it might not be obvious but it's also not possible for astexplorer to explain every option for every parser. And I also understand that having a "top level configuration" might be easier, but I also don't want to add something for every possible combination of settings.
Given that it's already possible to get the output you want, and the fact that I'm currently trying to change how parsers work (which means I'm not sure how something like recast which accepts another parser as input will work), makes me rather not want to add this.
That doesn't mean that something like this won't be added in the future, but I don't want to add it right now.

arnebech:
> I understand that it might not be obvious but it's also not possible for astexplorer to explain every option for every parser. And I also understand that having a "top level configuration" might be easier, but I also don't want to add something for every possible combination of settings.

Yeah, definitely a balance of what options to expose and how. I'd think that parsing a tsx file is pretty common task, so would be great to see some higher level support for that.

In my case we use jscodeshift with the 'tsx' parser option and wanted to find a good tool to explore the tsx ast tree. 

In any case, thanks for you work on this (and jscodeshift)!

PR #566, opened 2021-02-12
Add ERML parser

obadakhalili:
ERML is a super easy-to-write DSL that represents conceptual data models using the Entity-Relationship modeling concepts. You can read more about it from the [docs](https://erml.netlify.app/).


obadakhalili:
Does it usually take that long before answering a PR?

PR #572, opened 2021-03-01
Babel 7.13

JLHwung:
Updated `@babel/parser` to 7.13
Added new plugins supported in `@babel/parser`.


fkling:
I had to run `yarn upgrade babylon7` as well to upgrade `@babel/parser` to 7.13. Should be deployed in a bit.

PR #573, opened 2021-03-06
feat: add recast as transformer as well

### LINKED ISSUE #131: Add Recast as transformer as well
This would allow to creating pure recast transforms. I could imagine the module to export a single function which gets passed a reference to recast, the source and maybe even the ast parsed by the currently selected parser. The function can perform changes to the AST and uses recast's `print` method to return the changed source.

Zzzen:
close #131


fkling:
I'm a little bit worried that it might be confusing that you can select one parser via the settings and another parser in the transform itself.
However, that's because of a limitation of astexplorer, which already confuses people so maybe it doesn't matter.

Thank you :)

PR #574, opened 2021-03-16
feat: implement a typescript transformer

kevinramharak:
This PR will implement the typescript transformer API. The existing infrastructure made it extremly easy to implement.

### DIFF
--- a/website/src/parsers/js/transformers/typescript/codeExample.txt
+++ b/website/src/parsers/js/transformers/typescript/codeExample.txt
@@ -1,0 +1,32 @@
+/**
+ * The transformer signature is based on https://github.com/cevek/ttypescript#program
+ */
+export default function (program) {
+  const checker = program.getTypeChecker();
+  return (context) => {
+    return (sourceFile) => {
+      const visitor = (node) => {
+        // This branch evaluates '2 + 2' like expressions and replaces the node with the result (in this case '4')
+        if (ts.isBinaryExpression(node)) {
+          if (ts.isNumericLiteral(node.left) && ts.isNumericLiteral(node.right)) {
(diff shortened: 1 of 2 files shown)


kevinramharak:
NOTE: it does use some internal API's:
- The `ts.Writer` interface and `ts.createTextWriter` factory function
- The `ts.Printer` methods that write to this interface

There are no public API's available that implement these, and reimplementing them (copy & pasting) them seems like a wasted effort.

kevinramharak:
@fkling Any feedback or comments? If not then can this be merged?

fkling:
Mmh, I thought I has merged this in the meantime, but I guess not.

I just hope it's not going to be difficult to keep this maintained :)

kevinramharak:
Thanks for the merge. Works well on production. I doubt the code will break anytime soon, im just taking a shortcut in the API to not have to run the whole bloated pipeline for a small piece of code. I'll take another look to see if I can find equivelent public api calls, but the emit api is a bit of a pain to work with to just run a single transformer on a piece of code.

The code using the internal API are these lines/statements:
```ts
    // createTextWriter(newline: string)
    const writer = ts.createTextWriter(host.getNewLine());

    // ...

    // printer is part of the public API, writeFile is marked as internal
    // writeFile(sourceFile: ts.SourceFile, writer: ts.EmitTextWriter, ...optional stuff)
    printer.writeFile(resultFile, writer, void 0);
    
    // technically these are part of the internal API, because they are dependent on the createTextWriter call
    const result = writer.getText();
    writer.clear();
```

kevinramharak:
@fkling Turns out the public API does exactly what I did but its name and type declarations confused me. So the PR #579 should remove your concerns. Those calls are all part of the public api so it should address your concerns in maintainability.

PR #579, opened 2021-04-26
fix: remove internal typescript api calls in the typescript transformer

kevinramharak:
Changes the typescript transformer to use the public api instead of internal typescript API calls.
Also:
 - added a type check to give some sane feedback if the user messes up.
 - added some comments to document what each piece of code does.


kevinramharak:
@fkling Cold you merge this? This should adress the concerns about maintainability.

fkling:
Thank you!

PR #584, opened 2021-05-22
provide location information for shift-parser

bakkot:
I don't love the global state approach to storing side tables, but it seems to be the preferred approach (I copied it from the [typescript parser](https://github.com/fkling/astexplorer/blob/a565b32ecce73118ad82bbf5997367b19bd02f94/website/src/parsers/js/typescript.js#L66)).


fkling:
I don't particularly like it either, but it's the simplest way for now. With the changes that might come one day the approach has to change anyway (basically the AST will need to be traversed once to add "annotations").

PR #586, opened 2021-06-03
feat: [transformers] add example formatting + compatible parsers

ehzhang:
Looking for a little discussion on approach! I use AST explorer only every so often and most recently I found myself stumbling a bit again when I tried to write a jscodeshift transform, and was only met with the error below:

image

Here's an attempt to demystify some of this by adding the `module.exports.parser` directly to the code example (with a cute little formatter to automatically inject the "tsx" transformer)

Here's a gif of the flow:

https://user-images.githubusercontent.com/3484848/120608590-34d2a500-c406-11eb-86da-6c49c89bc298.mp4

----

This adds two new configuration options for transformers:
- compatibleParserIDs
- formatCodeExample

compatibleParserIDs is used in SELECT_TRANSFORMER as an alternative to
always resetting the parser when changing the transformer or turning it
on. If we are already on a compatible parser, we won't change.

formatCodeExample supports custom formatting/templating for code
examples - in the case of jscodeshift, this is used to set the
module.exports.parser based on current parser.

In combination, this supports a workflows like:
- Copy in a typescript file contents
- Change the parser to typescript
- Turn on the transform for jscodeshift

### DIFF
--- a/website/src/parsers/js/transformers/jscodeshift/codeExample.txt
+++ b/website/src/parsers/js/transformers/jscodeshift/codeExample.txt
@@ -1,3 +1,7 @@
+// jscodeshift can take a parser, like "babel", "babylon", "flow", "ts", or "tsx"
+// Read more: https://github.com/facebook/jscodeshift#parser
+export const parser = '{{parser}}'
+
 // Press ctrl+space for code completion
 export default function transformer(file, api) {
   const j = api.jscodeshift;
--- a/website/src/parsers/js/transformers/jscodeshift/index.js
+++ b/website/src/parsers/js/transformers/jscodeshift/index.js
@@ -5,6 +5,20 @@ const ID = 'jscodeshift';
 
 const sessionMethods = new Set();
 
+// https://github.com/facebook/jscodeshift#parser
+const getJscodeshiftParser = (parser, parserSettings) => {
+  if (parser === 'typescript') {
+    if (parserSettings.typescript && parserSettings.typescript.jsx === false) {
+      return 'ts'
+    }
+    return 'tsx'
+  }
+  if (parser === 'flow') {
+    return 'flow'
+  }
+  return 'babel'
+}
+
 export default {
   id: ID,
   displayName: ID,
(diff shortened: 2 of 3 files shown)


website/src/parsers/js/transformers/jscodeshift/index.js
```diff
@@ -5,13 +5,35 @@ const ID = 'jscodeshift';
...
+      return 'ts'
+    }
+    return 'tsx'
+  }
+  if (parser === 'flow') {
+    return 'flow'
+  }
+  return 'babel'
+}
+
```
ehzhang, line 21: there's probably some larger question about maintainability here since there are quite a few parsers that are capable of jsx + typescript these days beyond just the typescript parser, like @babel/parser, @typescript-parser/eslint, etc

I called out just the typescript parser as the primary one here - but alternatively, if a parser isn't defined here then we'll at least have the module.exports.parser so it'll be a bit more obvious what's needed to get jscodeshift less upset 🤔

ehzhang: 🤷 Another alternative is to just downscope this entire feature to just adding the `module.parser` to the codeExample and call it there.

fkling:
I think this is a good solution for now, even if it's just for jscodeshift.

In the long term I want parsers and transformers to be more integrated. The current isolation of parsers and transformers is causing confusion and also limits how transformers can be used.
Instead I want to make the transformer directly selectable and configurable, including the parsers it supports and should use.

website/src/parsers/js/transformers/jscodeshift/codeExample.txt
```diff
@@ -1,3 +1,7 @@
+// jscodeshift can take a parser, like "babel", "babylon", "flow", "ts", or "tsx"
+// Read more: https://github.com/facebook/jscodeshift#parser
+module.exports.parser = {{parser}}
```
fkling, line 3: Given that we use ES module syntax, should should rather be

```
export const parser = {{parser}}
```

although I would probably even wrap the placeholder into quotes and at least make syntactically valid JavaScript:

```
export const parser = '{{parser}}';
```

ehzhang: thanks for the suggestion! i updated the commit with these changes/

ehzhang:
Hi @fkling! Wanted to follow up to see if there was anything else you'd like me to address. Thanks! :)

fkling:
Sorry, just not finding much time recently.

PR #587, opened 2021-06-03
Chore: upgrade vue-eslint-parser

ota-meshi:
This PR upgrades vue-eslint-parser and adds new options.

### DIFF
--- a/website/src/parsers/vue/vue-eslint-parser.js
+++ b/website/src/parsers/vue/vue-eslint-parser.js
@@ -18,6 +18,9 @@ export default {
   },
 
   parse(parser, code, options) {
+    if (Object.keys(options).length === 0) {
+      options = this.getDefaultOptions();
+    }
     return parser.parse(code, options);
   },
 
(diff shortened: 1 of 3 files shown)


website/src/parsers/vue/vue-eslint-parser.js
```diff
@@ -18,6 +18,9 @@ export default {
   },
 
   parse(parser, code, options) {
+    if (Object.keys(options).length === 0) {
+      options = this.getDefaultOptions();
+    }
```
fkling, line 23: I don't think this is strictly necessary (the default options are always passed), but there might be some edge case I'm not considering. Did you add this because you ran into issues?

ota-meshi: I added this because I had the same issue as acorn.

https://github.com/fkling/astexplorer/pull/588#discussion_r647355376

fkling: I see. I doesn't hurt to add this, I just hope it won't be necessary in the future :)

PR #588, opened 2021-06-04
Chore: upgrade acorn

ota-meshi:
This PR upgrades acorn and adds support for ES2022.


website/src/parsers/js/acorn.js
```diff
@@ -24,6 +24,9 @@ export default {
   },
 
   parse(parsers, code, options={}) {
+    if (Object.keys(options).length === 0) {
+      options = this.getDefaultOptions();
+    }
```
fkling, line 29: Same question as in the other MR: Did you add this because you ran into issues?

ota-meshi: After pressing the "reset" button, `{}` was passed. So I added this branch. The acorn parser crashes with an error even before I make my changes.

image

image


I referred to another parser.
https://github.com/fkling/astexplorer/blob/17d7bf5e628a2d3e3ca4f7cec4a14d531811c881/website/src/parsers/regexp/regexpp.js#L26

PR #589, opened 2021-06-08
Fix node range handling in Hermes parser adapter

motiz88:
Quick follow up from #504 fixing location handling with the Hermes parser. Adds a `nodeToRange` method supporting the two different AST formats generated by `hermes-parser`, as well as matching `locationProps`.

Test plan:

1. Ensure that "Hide location data" works for both `babel: false` and `babel: true`.
2. Ensure that hovering a node highlights the corresponding code for both `babel: false` and `babel: true`.

### DIFF
--- a/website/src/parsers/js/hermes.js
+++ b/website/src/parsers/js/hermes.js
@@ -30,7 +30,7 @@ export default {
   displayName: pkg.name,
   version: pkg.version,
   homepage: pkg.homepage || 'https://hermesengine.dev/',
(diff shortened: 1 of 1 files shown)


motiz88:
CI seems to be having unrelated difficulties, but for some reason I can't login to restart the job.

motiz88:
@fkling friendly ping! Sorry for missing this issue in the initial Hermes PR.

fkling:
No problem :)

PR #598, opened 2021-08-03
feat: Add stylis

jantimon:
[stylis](https://github.com/thysultan/stylis.js) is "A Light–weight CSS Preprocessor." used by libraries like [styled-components](https://github.com/styled-components/styled-components/blob/d96101a102b59a8f9ac4b6f278594674897282bb/packages/styled-components/package.json#L76)

this pr adds support for parsing and manipulation of css using stylis:

stylis

### DIFF
--- a/website/src/parsers/css/stylis.js
+++ b/website/src/parsers/css/stylis.js
@@ -1,0 +1,61 @@
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'stylis/package.json';
+
+const ID = 'stylis';
+
+function removeLength(node, lineLengths) {
+  if (!node || typeof node !== 'object') return;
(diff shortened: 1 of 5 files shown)


website/src/parsers/css/stylis.js
```diff
@@ -0,0 +1,61 @@
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'stylis/package.json';
+
+const ID = 'stylis';
+
+function removeLength(node, lineLengths) {
```
fkling, line 6: What's the purpose of this?

jantimon: It's a workaround - if `length:0` is present astexplorer will hide all children

PR #599, opened 2021-08-11
Update to Babel 7.15

nicolo-ribaudo:
This also adds an option to use the new "Hack" pipeline proposal.

Technically it replaces the "Smart" proposal, but since it's still supported by Babel I didn't remove the option.


website/src/parsers/js/babylon7.js
```diff
@@ -93,20 +100,27 @@ export default {
...
+            proposal: pipelineOptions.proposal,
+            topicToken: pipelineOptions.hackTopicToken,
+          }];
         case 'recordAndTuple':
           return ['recordAndTuple', { syntaxType: 'hash' }];
         default:
           return plugin;
       }
     });
+    console.log(options);
```
JLHwung, line 123:
```suggestion
```

JLHwung (approved):

website/src/parsers/js/babylon7.js
nicolo-ribaudo: Whops

nicolo-ribaudo:
Thanks for merging!

PR #602, opened 2021-08-20
(feat) - use swc and upgrade to webpack v5

JoviDeCroock:
SWC uses `import.meta.url` which is only supported in Webpack 5, this is the reason behind upgrading the deps inside of this PR as well. Currently seeing some stylistic issues (icons missing) in the UI but apart from that everything is already building.

NOTE: currently the top-nav icons aren't working trying to figure out what's wrong there

CC @kdy1 it looks like the wasm implementation is having some issues parsing the options, is this expected, I took this code from the example [here](https://github.com/swc-project/swc/tree/master/wasm/example)

```
failed to parse options

Caused by:
    invalid type: null, expected struct ParseOptions at line 1 column 4
```

When using this code

```js
  transform(swc, options, code) {
    try {
      const Plugin = eval(`
        import Visitor from "@swc/core/Visitor";
        ${options}
      `);

      return swc.transformSync(code, {
        "jsc": {
          "target": "es2016",
          "parser": {
            "syntax": "ecmascript",
            "jsx": true,
            "dynamicImport": true,
          }
        },
        plugin: m => new Plugin().visitProgram(m)
      });
    } catch (e) {
      throw new SyntaxError(e);
    }
  },
```

Also should `wasm` export the `Visitor` function?


kdy1:
No, it's not expected.
But `@swc/wasm` does not support `plugin`.
Can you try it without `plugin`?

JoviDeCroock:
Oh 😅 even without plugin it fails though in the [`parseSync`](https://github.com/fkling/astexplorer/pull/602/files#diff-d577505d3c06ca68c06a3edb0b6dabe22aa5987caf3f2085c241f7e524bd6b03R20) call already (syntax: 'ecmascript').

Hmm, so transform wouldn't be possible with the wasm output

kdy1:
Actually swc has repl based on `@swc/wasm`.
https://swc.rs/repl/

I could compile with your `.swcrc`

https://swc.rs/repl/#%7B%22rawCode%22%3A%22%5Cn%20%20class%20Project%20%7B%5Cn%20%20%20%20constructor(name)%20%7B%5Cn%20%20%20%20%20%20this.name%20%3D%20name%5Cn%20%20%20%20%7D%5Cn%20%20%7D%3B%5Cn%5Cn%20%20export%20class%20SwcProject%20extends%20Project%20%7B%5Cn%20%20%20%20constructor(name%2C%20version)%20%7B%5Cn%20%20%20%20%20%20super(name)%3B%5Cn%20%20%20%20%20%20this.version%20%3D%20'20210629'%3B%5Cn%20%20%20%20%7D%5Cn%5Cn%20%20%20%20showOff()%20%7B%5Cn%20%20%20%20%20%20console.log('swc%20is%20a%20useful%20project')%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%22%2C%22rawConfig%22%3A%22%7B%5Cn%5C%22jsc%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%5C%22target%5C%22%3A%20%5C%22es2016%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%5C%22parser%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22syntax%5C%22%3A%20%5C%22ecmascript%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22jsx%5C%22%3A%20true%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22dynamicImport%5C%22%3A%20true%5Cn%20%20%20%20%20%20%20%20%20%20%7D%5Cn%20%20%20%20%20%20%20%20%7D%5Cn%7D%22%7D

JoviDeCroock:
I might be doing something wrong, I'll look for that code, thank you for the pointers!

JoviDeCroock:
Okay, the parsing is working now, do you think enabling plugins would be on the roadmap for the webassembly version? I think a tool like astexplorer is immensely useful to create plugins

kdy1:
No, I'm going to provide rust-based plugin api and deprecate `print` and `parse`.
Currently, I don't have any plan related to wasm plugins.

JoviDeCroock:
Hmm in that case it might not be worth adding this functionality to `astexplorer` as `parse` gets deprecated 😅

kdy1:
Oh, I missed one thing.
I think `parse` should be preserved to help plugin developers.

JoviDeCroock:
Oh awesome, thank you so much for taking time out of your day to chime in here! It's immensely appreciated, your work on swc is so awesome to see being pushed forward!

PR #608, opened 2021-09-18
fix: site stuck in infinite loop

shaharkazaz:
This PR fixes the following issues:

1. When using the Angular HTML parser, the site crashes (stuck in an infinite loop) with the following template:

```html
<div *ngFor="let hero of heroes; let i=index; let odd=odd; trackBy: trackById">
  ({{i}}) {{hero.name}}
</div>
```

This is caused by the `fixSpan` method, which isn't handling bound attributes from structural directives and has a `while` loop with a stop condition that doesn't meet in this case.

2. The attribute span indexes are incorrect since it returns the first `=` index after the attribute, for example:

```html
<div *ngFor="let hero of heroes; let i=index; let odd=odd; trackBy: trackById" [class.odd]="odd">
  ({{i}}) {{hero.name}}
</div>
```
* The `ngForOf` bound attribute should have the following span: `{start: 22, end: 28}` but it actually resolves to 
`{start: 39, end: 45}` which is the `=` char index of the template variable `i`.
* The `ngForTrackBy` bound attribute should have the following span: `{start: 59, end: 68}` but it actually resolves to `{start: 92, end: 102}` which is the `=` char of the `class.odd` bound attribute.

3. The input span indexes are wrong for inputs with structural directives (`[class.odd]="odd"`)

The span for this attribute should be `{start: 92, end: 95}` but it's actually `{start: 184, end: 187}` this is because this attribute is referenced both as an input to the `Template` node and as an input to the `Element` node (the `div`)

### DIFF
--- a/website/src/parsers/html/angular.js
+++ b/website/src/parsers/html/angular.js
@@ -83,6 +83,7 @@ function getNodeCtor(node) {
  *                  ^^^^^^^^^^ sub AST { start: 13, end: 23 }
  */
 function fixSpan(ast, code) {
+  const fixed = new Set();
   const KEEP_VISIT = 1;
   function visitTarget(value, isTarget, fn, parent) {
     if (value !== null && typeof value === 'object') {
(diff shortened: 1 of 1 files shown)


shaharkazaz:
@fkling Can you please review? 🙂

shaharkazaz:
@fkling Can this be reviewed? still crashes the site

fkling:
Thank you for taking the time to fix this @shaharkazaz! I apologize for being negligent with keep astexplorer up-to-date...
Since I'm not familiar with the Angular parser there is not much for me to review, I trust that this is the right thing to do. It certainly fixes the issues you pointed out.
Thank you again!

PR #624, opened 2021-12-29
Upgrade svelte transformer to version 3.44.3

metonym:
This PR upgrades the svelte transformer from version 3.20.1 to 3.44.3.

This allows for newer Svelte language features to be parseable.

For example, the [`$$slots` object](https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3250) was added in Svelte version 3.25. With v3.20, we encounter an "illegal variable name" error:

Screen Shot 2021-12-29 at 1 07 43 PM

The syntax is valid when using the latest version:

Screen Shot 2021-12-29 at 1 07 54 PM

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -149,7 +149,7 @@
     "solidity-parser-diligence": "^0.4.18",
     "source-map": "^0.6.1",
     "sqlite-parser": "^1.0.0-rc3",
-    "svelte": "^3.4.1",
+    "svelte": "^3.44.3",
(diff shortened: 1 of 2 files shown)


G-Rath (approved):

metonym:
Bumped svelte again to v3.46.2 to support bleeding edge features like`@const` tags and `style:prop` directives.

fkling:
Thank you!

PR #630, opened 2022-01-30
Allow JSX with @babel/eslint-parser

karlhorky:
Hi there! First of all, thanks so much for AST explorer, it's amazing for learning about ASTs and getting into things like ESLint plugin development.

I wanted to add the Babel preset `@babel/preset-react` in order to enable JSX parsing with `@babel/eslint-parser`.

For example, at [this demo](https://astexplorer.net/#/gist/04959b13b479ef26f67ef13e7bc657f0/ded84c9cf4a503832e0946d4182d4eb2828651d7) you see this:

Screen Shot 2022-01-30 at 18 55 22

```
unknown: Support for the experimental syntax 'jsx' isn't currently enabled (1:1):

> 1 | <a
    | ^
  2 |   rel={"asdf"}
  3 | />

Add @babel/preset-react (https://git.io/JfeDR) to the 'presets' section of your Babel config to enable transformation.
If you want to leave it as-is, add @babel/plugin-syntax-jsx (https://git.io/vb4yA) to the 'plugins' section to enable parsing.
```

---

However, I'm unfamiliar with the internals of the codebase here, so it's possible that I've gotten something wrong.


karlhorky:
Hm, I realize that in the `parse` method I don't actually need access to the preset, so I've simplified here (again, I'm not sure if I'm doing the right thing): 

[`ab8798f` (#630)](https://github.com/fkling/astexplorer/pull/630/commits/ab8798f7f536f4af31cf26f47f30416b8088afbb)

karlhorky:
Also moved `@babel/parser-react` from `devDependencies` to `dependencies` in [`9862e65`](https://github.com/fkling/astexplorer/pull/630/commits/9862e65b9a3249d2e3fe336ccf5d3efe3db06ea5), to indicate that it is a dependency required by a parser.

karlhorky:
I'm fine with switching to `@babel/plugin-syntax-jsx` if you'd prefer! I'm not sure what else `@babel/preset-react` adds...

fkling:
It includes two transform plugins that are not needed in this case: https://babeljs.io/docs/en/babel-preset-react . They are probably never invoked, but I guess that they would still be bundled.

karlhorky:
Ok no problem, switched to `@babel/plugin-syntax-jsx`.

Do you know if it's actually necessary to `require` the plugin in `loadParser`? I `require`d it just in case it was necessary, but maybe Babel is loading it by itself anyway?

website/src/parsers/js/babel-eslint-parser.js
```diff
@@ -13,13 +13,16 @@ export default {
...
-    require(['@babel/eslint-parser'], callback);
+    require(['@babel/eslint-parser', '@babel/plugin-syntax-jsx'], callback);
   },
 
   parse(parser, code) {
     const opts = {
       sourceType: 'module',
       requireConfigFile: false,
+      babelOptions: {
+        presets: ['@babel/plugin-syntax-jsx'],
```
fkling, line 24: > Is it necessary to require the plugin in loadParser? I required it just in case it was necessary, but maybe Babel is loading it by itself anyway just with the string from the config?

Have you tested it? (that's not a trick question 😆). I don't know what exactly babel does to resolve presets by strings (I guess we could find out), but I wouldn't expect this to work (usually those strings would be resolved to something on disk). But maybe babel is doing it in a way that's compatible with webpack... still I would probably pass the reference to the plugin to the options instead of referring to it by name.

However, having set all that and looking at how we configure things for the other babel parsers, it made me realize that all of this is unnecessary. It should work by directly enabling the JSX plugin in the parser without requiring any additional modules (because JSX support is built-into the parser; in fact, all that the syntax plugin does is [setting that option](https://github.com/babel/babel/blob/b91c3e15738ca9a390135bdce9d845da90c40733/packages/babel-plugin-syntax-jsx/src/index.ts#L20)):

```js
const opts = {
  // ...
  babelOptions: {
    parserOpts: {
      plugins: ['jsx'],
    }
  },
};
```

See https://babeljs.io/docs/en/options#parseropts and https://babeljs.io/docs/en/babel-parser#language-extensions

fkling (changes requested):

website/src/parsers/js/babel-eslint-parser.js
karlhorky: > still I would probably pass the reference to the plugin to the options instead of referring to it by name

Yeah I wanted to do this at the beginning, but after doing a bit of research, I couldn't see any examples where I could pass a `require`d plugin to the config.

karlhorky: > Have you tested it?

Nope, haven't even cloned the repo - just doing this all by manually editing files on GitHub.

karlhorky: Good point about the `parserOpts` though, that could be also a way to go. I'll make the change.

fkling:
I tested it locally and it seems to work fine now 👍🏻 Thank you!

karlhorky:
Great, thanks for testing and merge! 🙌 Tweeted about it haha https://twitter.com/karlhorky/status/1490662541462224897

PR #556, opened 2020-12-16
[WIP] Add swc JavaScript parser

punkeel:
> swc is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript.

Expose the swc parser in AST Explorer. This parser is built using Rust, and the client side version relies on WebAssembly.

Known bug: any action (editing the code or changing a setting) breaks the Autofocus feature, and requires a page reload.

Closes #555.

Issue #555, opened 2020-12-16
feature request: support swc


punkeel:
> swc is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript.

swc is a serious project with 10k stars on [GitHub](https://github.com/swc-project/swc). It is used in [Deno](https://github.com/denoland/deno) to transpile TypeScript code into JS. I'm not 100% sure, but there may be a WebAssembly version exposing parse / codegen functions.

It would be great to add support for swc in astexplorer.

cc @kdy1

RReverser:
> I'm not 100% sure, but there may be a WebAssembly version exposing parse / codegen functions.

It seems to have it right in the tree you provided: https://github.com/swc-project/swc/tree/master/wasm. If you can find or publish an npm package for bindings (see my https://github.com/RReverser/astexplorer-syn for an example), it should be fairly straightforward to integrate into ASTExplorer.

punkeel:
Indeed, `swc/wasm` exists and is deployed to npm. It's targeting nodejs, so I've opened https://github.com/swc-project/swc/issues/1277 to add a web version of it. I'll try playing a bit with AST Explorer to see if I can add support there, thanks for the syn pointer!

punkeel:
I've pushed a WIP pr #556 that has the following known issues:

1. It doesn't build, as `@swc/wasm` isn't meant to be used from browsers. To test it locally, I've manually built a browser-friendly version and replaced the `node_modules/...` files. See issue linked above in swc project.
2. When the code (or a setting) is changed, Autofocus is broken. Would you have any ideas what could cause this?

Thinking it could be WebAssembly shenanigans, I've tried wrapping the `parseSync` call in `JSON.parse(JSON.stringify(...))`, but it didn't help.

Aside from that, it was really easy to integrate in AST Explorer, ✨  magic ✨ !
<img width="1663" alt="swc astexplorer" src="https://user-images.githubusercontent.com/1859135/102410630-18c27f00-3ff1-11eb-8a69-bd59793b1af6.png">

RReverser:
> Aside from that, it was really easy to integrate in AST Explorer, ✨ magic ✨ !

Glad you liked it :) I've set it up a long time ago, and sometimes felt bad about all the Webpack magic it's doing, but it does seem to make contributions easier.

punkeel:
A first version of `@swc/wasm-web` was published, but another release is required to fix a bug, see https://github.com/swc-project/swc/issues/1291.

[...]

PR #637, opened 2022-02-28
feat(parser): add swc support via @swc/wasm-web

kwonoj:
This PR is one additional attempt to #555 for adding swc's parser into ast explorer.

I am aware there are prior attempts like https://github.com/fkling/astexplorer/pull/556 or https://github.com/fkling/astexplorer/pull/602 and given that both have not landed, honestly not sure if this PR is feasible to land. Also, feel free to close if the original PR https://github.com/fkling/astexplorer/pull/556 should be updated instead.

This PR attempts to preserve existing build as much, instead apply some workaround to bend over `@swc/wasm-web` to work with webpack@4. swc's web target build is for non-bundler cases which webpack@4 doesn't support some of the syntax. Change workarounds it via manually injecting path to the wasm binary.

Other changes are quite straightforward to follow contributing guidelines Local testing seems most features are working, however there maybe some unknown things need to be fix forward.


fkling:
Thank you for taking a stab at this. I'll take a look at this more closely next week. It's unfortunate that workaround are necessary to make it work with webpack@v4, but upgrading to webpack@v5 is probably too big of an effort at the moment.

jantimon:
Right now we are developing SWC Plugins and would love to use ASTExplorer as a tool - are there any blockers for this PR which we could support?

fkling:
Everything looked fine. The website has been updated.

jantimon:
thanks felix 🎉

PR #649, opened 2022-06-01
[Updates] typescript v4.7.0 & @typescript-eslint/parser v5.27.0

Ayc0:
Since the v4, TS added new syntaxes (specially in this new v4.7.0), so this PR aims to update both TS and @typescript-eslint/parser to be able to support those new syntaxes (and to be up to date with the latest version of TS in general):
- `typeof y.z<w>`
- `withGeneric<T>`
- covariant and contravariant `in` and `out`

See https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/ & https://github.com/typescript-eslint/typescript-eslint/releases/tag/v5.27.0.

## typescript

Before | After
-|-
image `out` here is improperly considered as an `Identifier` | image but now it's an `OutModifier`

## @typescript-eslint/parser

Before | After
-|-
image Before it was crashing | image But now it can properly parse `out` as seen by the `out: true` in the metadata of the generic

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -54,7 +54,7 @@
     "@humanwhocodes/momoa": "^1.0.0",
     "@mdx-js/mdx": "^1.5.8",
     "@swc/wasm-web": "^1.2.146",
-    "@typescript-eslint/parser": "^5.0.0",
+    "@typescript-eslint/parser": "^5.27.0",
     "@vue/compiler-dom": "^3.0.0-rc.10",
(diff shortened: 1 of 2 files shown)


Ayc0:
As a follow-up, Babel should also be updated to be able to support these new syntaxes:

image

fkling:
Thank you! I need to resolve lock file merge conflicts (due to babel update), but that was easy. Thank you for testing the changes beforehand!

PR #654, opened 2022-06-13
feat: add support for mathjs AST

mattvague:
Adds support for ASTs generated by http://mathjs.org


fkling:
Nice! It doesn't look like the AST contains positioning information, so it can't be corrolated with the source code. It would be nice if this supported or enabled eventually.

mattvague:
@fkling Sweet, thanks! I've started on a PR adding positional information to the AST [here](https://github.com/josdejong/mathjs/pull/2615) but will need some help from the lib author. Will create a new PR here to add support once that's done

PR #656, opened 2022-06-17
Improve ESLint v8 example rule 

guilhermesimoes:
Fixes the current error that appears on the site when you delete the `${i}` from the example:

> Fixable rules must set the `meta.fixable` property to "code" or "whitespace".
> Occurred while linting <input>:18
> Rule: "astExplorerRule"

With this change it will be easier to modify the example rule to make fixes and suggestions.


fkling:
Thank you!

guilhermesimoes:
It's already live!

PR #663, opened 2022-08-08
Upgrade svelte to 3.52

metonym:
Currently, the Svelte compiler is unable to recognize `svelte:element` tags or `@const` tags in `#if` blocks. This PR upgrades svelte to the latest minor version.

In the AST Explorer, the Svelte compiler is unable to parse the following code. It works in the [Svelte REPL](https://svelte.dev/repl/dd9b63a367cc46139ab770341d45f2d5?version=3.49.0) when using the latest minor version.

```svelte
<!-- 3.48.0 -->
{#if true}
  {@const a = 44}
    {a}
{/if}

<!-- 3.47.0 -->
<svelte:element this="h6">
  h6 tag
</svelte:element>

```

astexplorer – svelte@3.46.2

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -154,7 +154,7 @@
     "solidity-parser-diligence": "^0.4.18",
     "source-map": "^0.6.1",
     "sqlite-parser": "^1.0.0-rc3",
-    "svelte": "^3.46.2",
+    "svelte": "^3.52.0",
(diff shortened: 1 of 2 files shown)


fkling:
I'm sorry for letting this sit so long :-/

PR #664, opened 2022-08-09
Add Svelte as a top-level category

metonym:
Currently, Svelte is a parser under the HTML category. Although Svelte markup is a superset of HTML, Svelte can be considered its own language[^1].

This PR adds Svelte as its own top-level category, which also makes it easier to locate.

[^1]: [what-is-svelte.md](https://gist.github.com/Rich-Harris/0f910048478c2a6505d1c32185b61934)

ast-explorer-svelte


fkling:
We should have done that from the beginning, thank you!

PR #669, opened 2022-09-12
Babel 7.19

JLHwung:
In this PR we update `@babel/parser` to 7.19 and support more plugin options.

### DIFF
--- a/website/src/parsers/js/babylon7.js
+++ b/website/src/parsers/js/babylon7.js
@@ -50,8 +50,11 @@ export const defaultOptions = {
     'functionBind',
     'importAssertions',
     'jsx',
+    'regexpUnicodeSets',
   ],
+  decoratorOptions: { version: "2022-03", decoratorsBeforeExport: false, allowCallParenthesized: true },
   pipelineOptions: { proposal: 'hack', hackTopicToken: '%' },
+  typescriptOptions: { dts: false, disallowAmbiguousJSXLike: false },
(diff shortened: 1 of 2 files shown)


website/src/parsers/js/babylon7.js
```diff
@@ -102,20 +124,18 @@ export default {
...
         case 'decorators':
-          return ['decorators', {decoratorsBeforeExport: false, version:"2021-12"}];
+          return ['decorators', decoratorOptions];
         case 'pipelineOperator':
           return ['pipelineOperator', {
             proposal: pipelineOptions.proposal,
             topicToken: pipelineOptions.hackTopicToken,
           }];
-        case 'recordAndTuple':
-          return ['recordAndTuple', { syntaxType: 'hash' }];
```
JLHwung, line 118: This is removed because in 7.19 Babel defaults `syntaxType` to be `hash`.

website/src/parsers/js/babylon7.js
```diff
@@ -50,8 +50,11 @@ export const defaultOptions = {
     'functionBind',
     'importAssertions',
     'jsx',
+    'regexpUnicodeSets',
   ],
+  decoratorOptions: { version: "2022-03", decoratorsBeforeExport: false, allowCallParenthesized: true },
```
JLHwung, line 55: Here we default to the latest decorator versions

PR #673, opened 2022-11-15
Add new SQL parser: sql-parser-cst

nene:
Here's a pull request to support new parser for SQL.

As it supports parsing multiple SQL dialects (though for now only two) I've included a config option to switch between the dialects, plus a few more settings.

PS. This is an awesome project. I've been happily using this for years. Big thanks for creating and maintaining it 👍


fkling:
Thank you @nene! It makes me happy to hear that you find astexplorer useful 😃

PR #674, opened 2022-11-18
add the treeVisualization of ast

raintoway:
by the treeVisualization，users can view the ast directly。


fkling:
Thank you @raintoway! This is great, I've always wanted a graphical representation of the tree. The timing is a bit unfortunate though because I'm in the middle of rewriting astexplorer in Svelte. I'll have to think about how to go about this...

raintoway:
> Thank you @raintoway! This is great, I've always wanted a graphical representation of the tree. The timing is a bit unfortunate though because I'm in the middle of rewriting astexplorer in Svelte. I'll have to think about how to go about this...

i am pleasure to join the new project because it is useful and interesting

PR #681, opened 2022-11-30
Upgrade sql-parser-cst from 0.5.1 to 0.8.1

nene:
This version brings BigQuery support, so I also added the new language to the dropdown in settings. Plus the syntax tree produced by sql-parser-cst has had some major changes between 0.5 and 0.8 versions.

Normally I would avoid throwing PR-s to AstExplorer for every little upgrade, but this one completely changes how table and column names are represented, making the syntax tree in 0.5 largely incompatible with the one in 0.8.


vsviridov:
It's on `0.17.1` with even more dialects supported.

PR #683, opened 2022-12-07
feat: add astx transformer

jedwards1211:
This is self-promotion, since I'm the author of [astx](https://github.com/codemodsquad/astx).  It's a sweet new JS/TS transformer for doing pattern-based structural search and replace.

Note: I made `transformerMiddleware` pass the `parser` and `parserSettings` to the `transform` function...the way it was being done didn't make a whole lot of sense, for instance with `jscodeshift` it templates the selected parser into the transform code at first, but you can still select a different parser afterwards in astexplorer, making the two out of sync.  To me it makes more sense for the transformer to always use the parser and settings selected in the workbench (if applicable?)

Thanks so much for all the work you've done on astexplorer!  Without it I may never have gotten into writing codemods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ developers who want to create AST transforms. In fact, following transformers
 are included so you can prototype your own plugins:
 
 - JavaScript
+  - [astx][]
   - [babel][] (v5, v6)
   - [ESLint][] (v1, v2, v3)
   - [jscodeshift][]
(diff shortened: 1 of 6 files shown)


jedwards1211:
Oh it doesn't automatically rerun the transformer when I select a different parser in the workbench.  Need to make it do that.

PR #689, opened 2022-12-31
Update MDX, support version 1 and 2, as well as the multiple ASTs generated

ChristianMurphy:
mdx version 1 and version 2 have different syntax trees, this PR allows both versions' syntax trees to be previewed.

In addition, mdx internally uses multiple syntax trees.

Version 1 has a markdown AST (mdast/remark) and an HTML AST (hast/rehype).
Version 2 has a markdown AST (mdast/remark), an HTML AST (hast/rehype), and a JavaScript AST (esast/recma).

This updates AST Explorer to have a preview for all version 1 and version 2 ASTs.
As well as updates transformers to allow for transforms on both version 1 and version 2.

This also adds support for syntax extensions, supported in MDX version 2; including GFM, Math, Directives, and more.

/cc @silvenon, @remcohaszing, @wooorm

---

### Screenshots

**mdx 2 preview with mdx 2 transformer**
Screenshot 2022-12-30 at 23-04-40 AST explorer

**parser list for mdx**
parser-list

**mdx 2 settings/plugins**
Screenshot 2022-12-30 at 23-04-17 AST explorer


website/yarn.lock
```diff
@@ -112,7 +113,16 @@
     source-map "^0.5.0"
     trim-right "^1.0.1"
 
-"@babel/generator@^7.19.0", "@babel/generator@^7.8.4":
+"@babel/generator@^7.12.5", "@babel/generator@^7.20.7":
+  version "7.20.7"
+  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.20.7.tgz#f8ef57c8242665c5929fe2e8d82ba75460187b4a"
```
remcohaszing, line 118: You may want to run [`yarn-deduplicate`](https://github.com/scinos/yarn-deduplicate).

remcohaszing (commented):
Thanks! I really love astexplorer, but I miss MDX2 support

website/src/parsers/mdx/mdx2recma.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import defaultParserInterface from '../utils/defaultParserInterface';
+import pkg from 'mdx2/package.json';
+
+const ID = 'mdx2-recma';
```
remcohaszing, line 5: I'd either use the names `mdast` / `hast` / `estree` or `remark` / `rehype` / `recma`, but not a mix of those.

wooorm: yeah, same, should be `estree` probably

wooorm:
Any reason the old AST is still around? Why not drop it?

remcohaszing:
I also wouldn't mind dropping the old AST

website/src/parsers/mdx/mdx2recma.js
ChristianMurphy: Good point, updated it to be named `esast` (https://github.com/syntax-tree/esast)

website/yarn.lock
ChristianMurphy: good idea, done

ChristianMurphy:
> Any reason the old AST is still around? Why not drop it?

A couple reasons.
1. there are still plugins out there for mdx 1, which could benefit from having a tool to help inspect ASTs
2. ASTExplorer has some precedence with keeping multiple major versions of parsers. For example:
   * [three major versions of the babylon parser](https://github.com/fkling/astexplorer/tree/ac2cfcc5a1fed0b7b7bea1af48f6840230576bf8/website/src/parsers/js)
   * [four major versions of babel eslint](https://github.com/fkling/astexplorer/tree/ac2cfcc5a1fed0b7b7bea1af48f6840230576bf8/website/src/parsers/js)
   * [five major versions of eslint itself](https://github.com/fkling/astexplorer/tree/ac2cfcc5a1fed0b7b7bea1af48f6840230576bf8/website/src/parsers/js/transformers)

I'd also be fine with dropping it, but defaulted to keeping it.

wooorm:
Re 2: Old versions seems to be around to support old snippet links, they seem to be hidden: https://github.com/fkling/astexplorer/commit/fad47c2a66c302f3a10361b11aad1a36703cd72d

Re 1: MDX 0 and 1 ASTs, IMO, are more the “absence” of an AST. It’s mdast, but replaces a `html` node with a `jsx` node, with a slightly different regex. And it adds `import`/`export` nodes, which are similarly strings, with different regexes.
I don’t think it’s very interesting to look at this AST. Now that there are nodes for things in v2, I do think it gets interesting, and we could start versioning to keep snippet links alive?

remcohaszing:
I suppose MDX 3 should be added by now too.

PR #705, opened 2023-12-23
[htmlparser2] add 'endIndex' to 'locationProps'

vassudanagunta:
Hide Location Data should hide this prop as well. (Note its use in nodeToRange)

### DIFF
--- a/website/src/parsers/html/htmlparser2.js
+++ b/website/src/parsers/html/htmlparser2.js
@@ -10,7 +10,7 @@ export default {
   displayName: ID,
   version: pkg.version,
   homepage: pkg.homepage || 'https://github.com/fb55/htmlparser2',
-  locationProps: new Set(['startIndex']),
+  locationProps: new Set(['startIndex', 'endIndex']),
   typeProps: new Set(['type', 'name']),
 
   loadParser(callback) {


vassudanagunta:
Felix, while I've got your attention: Would you use React and Expressjs if you were building ASTExplorer from scratch today? If not, what?

I'm building a playground app for some new tech I'm doing on the side. So far it's hand coded HTML, CSS and HTML. Coding is not my day job, I'm a relative amateur, so would prefer not to use a platform with a long/hard leaning curve or one that's gonna make my brain hurt (I'm not young any more).

I'm just very impressed with the elegance and power, yet simplicity of ASTExplorer so thought I'd ask you. :)

fkling:
Thank you @vassudanagunta!

> Would you use React and Expressjs if you were building ASTExplorer from scratch today? If not, what?

A lot has changed in the last ~9 years. Personally I've found using [Svelte](https://svelte.dev/)/[SvelteKit](https://kit.svelte.dev) delightful to use. From what others it seems to be easier to approach for people not as familiar with frontend development.

> I'm just very impressed with the elegance and power, yet simplicity of ASTExplorer so thought I'd ask you. :)

Thank you, it makes me happy to hear that :) That's kind of wait I'm aiming for :D

